PR #7, opened 2015-03-25
Few Fixes

hostianer:
Chage IP Host Adress and added the correct .htaccess changes.

### DIFF
--- a/source/developers-guide/vagrant-phpstorm/index.md
+++ b/source/developers-guide/vagrant-phpstorm/index.md
@@ -148,14 +148,15 @@ After you get the `"BUILD SUCCESSFUL"` message you can run the full build comman
     $ ant build-unit
 
 ### Step 3
-Now edit your .htaccess and change the Rewrite Path to /shopware.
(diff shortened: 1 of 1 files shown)


source/developers-guide/vagrant-phpstorm/index.md
```diff
@@ -146,15 +146,30 @@ Input fields:
 After you get the `"BUILD SUCCESSFUL"` message you can run the full build command.
 
     $ ant build-unit
-  
+
+### Step 3
```
bcremer: Is that still necessary? I've never used the `RewriteBase` for shopware installations on various setups. I pretty sure it works well on the vagrant machine without that step.

hostianer: You`re right. A few minutes go and wasnt working. Now it works also if you leave it as it is :1

bcremer:
Thank you @hostianer for your pull request.

Please create separate Pull Requests for each semantic change so we can review and merge them separately without blocking each other.

hostianer:
@bcremer ok no problem will notice that for future pull requests.

PR #9, opened 2015-03-27
Fix: Error in createTextField()-function

t2oh4e:
The createTextField-function does not accept an array. The correct way is to enter the strings directly as parameters.


bcremer:
Thanks @t2oh4e

PR #18, opened 2015-04-08
Update documentation for unitize mixin

t2oh4e:
According to the unitize.less the documented way of using the unitize mixin is deprecated. Updated it to the new one.

Actually I don't know wether it is necessary to mention the base value of 16 here. Wouldn't it be better if there was an additional paragraph about the @remScaleFactor and how to change that if necessary?

### DIFF
--- a/source/designers-guide/theme-startup-guide/index.md
+++ b/source/designers-guide/theme-startup-guide/index.md
@@ -179,11 +179,11 @@ The following example shows how to use the mixin using a ```12px font-size```:
 
 ```
 p {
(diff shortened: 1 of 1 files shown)


bcremer:
@klarstil may you please review this.

klarstil:
Hey @t2oh4e, thanks for your PR. You're right, the documentation of the ```.unitize``` mixin was misleading. The previously documented way is still supported but we prefer the new one.

@bcremer The PR is fine and can be merged by you.

bcremer:
Thanks, was deployed right now.

PR #23, opened 2015-04-10
Add Styling guide

tiagojsag:
Initial styling guide, defining a basic rule set that future edits should follow, in order to ensure we have a consistent styling everywhere.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -1,0 +1,52 @@
+# Shopware Development Documentation Style Guide
+
+This document defines some rules that should be used when editing content on this project. They aim at providing the end user with a consistent experience when reading different pages/version inside this project.
+
+## Language
+
+All documents should be written in American or UK English. Should it be necessary, disambiguation between the two should be done on a per occurrence basis, giving preference to the one most commonly used on similar contexts (other documentation sites, for example)
+
+## Naming conventions
+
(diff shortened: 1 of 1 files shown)


bcremer:
I like the styling guide idea. I was also thinking about some kind of meta space where pages like this and "how to contribute" etc. are placed.

It does not fit in the existing "Sysadministrators/Developer/Designers" spaces we currently have. Any ideas where to put this page?

tiagojsag:
I would say that we add a link in the footer. some ideas for this section:
- How to contribute => link to github
- Credits => I think repos have a contributor list page somewhere
- Styling guide => my original idea was for this to NOT be in the final docs, as it's only relevant for authors, not for readers. But since we are adding this section, it makes sense to add this there

bcremer:
With "How to contribute" I meant "How to contribute to the documentation".

I like how dev-human.com does this;

http://dev-human.com/docs/contribute/
http://dev-human.com/docs/running-locally/

tiagojsag:
Yeah, me too. the link to github is a link to this repo, not the main SW repo.

I like the "contribute" page, but I honestly don't know if we need a very visible header section with 3 pages for it. I think I would prefer to have a general "how to contribute" page on the devdocs page, that would have some basic info about contributing, and then have the details here, on this repo. I think this separates better between "actual docs" and "contributing"

tiagojsag:
Will create a different PR

PR #25, opened 2015-04-14
Add contribution guide

tiagojsag:
Add contribution guide page, informing users how they can contribute to Shopware core code and to the devdocs

### DIFF
--- a/source/contribute/index.md
+++ b/source/contribute/index.md
@@ -1,0 +1,81 @@
+---
+layout: default
+title: Contribute
+github_link: contribute/index.html
+indexed: true
+---
+
+## Contribute to Shopware
+
+Shopware is available under dual license (AGPL v3 and proprietary license). If you want to contribute code (features or bugfixes), you have to create a pull request and include valid license information. You can either contribute your code under New BSD or MIT license.
+
+If you want to contribute to the backend part of Shopware, and your changes affect or are based on ExtJS code, they must be licensed under GPL V3, as per license requirements from Sencha Inc.
+
+If you are not sure which license to use, or want more details about available licensing or the contribution agreements we offer, you can contact us at <contact@shopware.com>.
+
+- **Contribution guide**: <https://github.com/shopware/shopware/blob/master/CONTRIBUTING.md>
+- **License**: Dual license AGPL v3 / Proprietary
+- **Github Repository**: <https://github.com/shopware/shopware>
+- **Issue Tracker**: <http://jira.shopware.de/jira>
+
+## Contribute to the Dev Docs
+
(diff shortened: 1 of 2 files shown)


STYLEGUIDE.md
```diff
@@ -0,0 +1,52 @@
+# Shopware Development Documentation Style Guide
```
bcremer: Can this file be removed? Seems to be duplicated by `source/contribute/index.md`

source/contribute/index.md
```diff
@@ -0,0 +1,78 @@
...
+- CSS
+- JavaScript
+- ID
+- HTML
+- HTTP/HTTPS
+- SSL
+- Responsive/Base (for themes names)
+- shopware AG (company name)
+
+When referring to specific parts of Shopware's interface, like buttons or modules, the naming should be an exact match to what users expect to find in the actual code/interface. Examples:
```
bcremer: Please remove that paragraph for now. Especially the Storefront vs. Frontend and Product vs. Article is quite controversial.

source/contribute/index.md
```diff
@@ -0,0 +1,78 @@
...
+- **Github Repository**: <https://github.com/shopware/devdocs>
+- **License**: Proprietary
+
+Contributions to this project should aim at improving the overall coverage / quality of the Shopware 5 documentation page.
+ 
+All contributions should respect the following styling guidelines.
+
+### Language
+
+All documents should be written in American or UK English. Should it be necessary, disambiguation between the two should be done on a per occurrence basis, giving preference to the one most commonly used on similar contexts (other documentation sites, for example)
```
bcremer: Regarding American vs. UK English: In my opinion we should go with only one of them.
I highly opt for American English because most programming languages/markups use american english name.
See CSS `color` for example.

tiagojsag: ok, I'll update it

source/contribute/index.md
tiagojsag: And that's why we need a rule for it. Otherwise we will have "articles" on half of the code/docs, and "products" on the other half. Loss in consistency means potential problems for new users who are not so familiar with Shopware or our double naming. I vote for keeping this

STYLEGUIDE.md
tiagojsag: done

source/contribute/index.md
bcremer: As written before. It's controversial inside shopware and we should discuss it first.

tiagojsag:
@bcremer I still prefer we have a coherent approach across the docs in the Articles/products and frontend/storefront topic, so I added them as "recommended" to the docs

PR #27, opened 2015-04-16
php_memory size

hostianer:
changed php_memory size from 128M to 256M


bcremer:
Thank you for your contribution @hostianer.

PR #68, opened 2015-06-08
Added JavaScript to "Using CSS files in themes" guide

tobiasberge:
@klarstil would you please review the guide

### DIFF
--- a/source/designers-guide/css-file-usage/index.md
+++ b/source/designers-guide/css-file-usage/index.md
@@ -1,25 +1,41 @@
 ---
 layout: default
-title: Using CSS files in themes.
+title: Using CSS and JavaScript files in themes.
 github_link: designers-guide/css-file-usage/index.md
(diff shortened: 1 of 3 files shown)


source/designers-guide/css-file-usage/index.md
```diff
@@ -1,25 +1,41 @@
...
+    └── _public
+        └── src
+            ├── css
+            │   └── example.css
+            └── js
+                └── example.js
 ```
 
-As the second step you will have to define the CSS files you would like to use inside your custom theme. This can be done by adding an array to your `Theme.php` file that contains the specific file paths, as the following example shows:
+As the second step you will have to define the CSS or JavaScript files you would like to use inside your custom theme. This can be done by adding an array to your `Theme.php` file that contains the specific file paths, as the following examples show:
```
tiagojsag: show => shows

PR #73, opened 2015-06-10
Add new guide to extend services, add themes and overwrite themes tem…

OliverSkroblin:
…plates


tiagojsag:
@bcremer I finished my review, would appreciate if a third set of eyes could take a look at the final result before merging.

bcremer:
@OliverSkroblin need to further review the content. I want to avoid duplication with https://developers.shopware.com/developers-guide/plugin-quick-start/.

PR #75, opened 2015-06-12
Use merge function for simple updates

hlohaus:


tiagojsag:
Update also the demo code.

PR #77, opened 2015-06-17
5.0.2 - Document snippet showSnippetPlaceholder option

tiagojsag:
Document the new showSnippetPlaceholder option.
Do not merge before the release of SW 5.0.2


bcremer:
We need to find a way to document the minimal shopware release inside of the documentation. If people stumble upon the `showSnippetPlaceholder` and try it in 5.0.1 they will be dissatisfied.

```
  'snippet' => array(
         'readFromDb' => true,
         'writeToDb' => false,
         'readFromIni' => true,
         'writeToIni' => true,
        'showSnippetPlaceholder' => true // available since SW 5.0.2
     ),
```

PR #80, opened 2015-06-18
added Plugin to expand/collapse large code tags.

arnoldstoba:


tiagojsag:
Functionally it works as intended.
@klarstil can you take a look at this please?

bcremer:
@arnoldstoba @tiagojsag can you please leave a usage example as a comment here?

tiagojsag:
http://gyazo.com/b200a6df84aab06bcbe74ed867cea043
Currently is used automatically in all code blocks that exceed a certain preconfigured height

bcremer:
OK tested this feature and I generally like it.

- Can we please increase `maxHeight` to 350px or so.
- The expand "Button" is a bit bulky and uses value space for it self. can we find another solution for the button. eg. float on the right bottom of the code box with 20% width or so.

klarstil:
@arnoldstoba I checked the code and the style. It looks good for me so far.

I agree with @bcremer. The button is way too big, we should consider another solution

arnoldstoba:
Updated the button style and maxHeight (to 370px) in the latest commit.
http://gyazo.com/922f5111252e8ec52cc106ce7b9c8227

PR #92, opened 2015-07-16
added missing notice for removed `s_articles_details.impressions`

uehler:


tiagojsag:
Thanks for the contribution @uehler

PR #102, opened 2015-08-03
Add service example

dnoegel:


exampleplugins/Frontend/SwagService/Bootstrap.php
```diff
@@ -0,0 +1,95 @@
...
+ */
+class Shopware_Plugins_Frontend_SwagService_Bootstrap extends Shopware_Components_Plugin_Bootstrap
+{
+    /**
+     * Register the namespace of our plugin globally
+     */
+    public function afterInit()
+    {
+        $this->Application()->Loader()->registerNamespace(
+            'Shopware\SwagService',
```
bcremer: Can you `ShopwarePlugins\SwagService` so it is consistent with the other example plugins.

exampleplugins/Frontend/SwagService/Bootstrap.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Return an instance of our new service class
+     *
+     * @return \Shopware\SwagService\Component\TaxCalculator
+     */
+    public function onInitTaxCalculator()
+    {
+        return new \Shopware\SwagService\Component\TaxCalculator(
+            $this->get('corelogger')
```
bcremer: We should use `pluginlogger` here. The `corelogger` should be reserved for logging within the shopware core.

tiagojsag:
I just fixed a couple of minor typos, nothing special. @dnoegel you will be a native speaker soon (so you can stop using that german language thing) :P

dnoegel:
Thanks @tiagojsag for your fixes. @bcremer: Changed plugin namespace

PR #108, opened 2015-08-12
Link fix

kayyyy:
The customerID Original Object links to `https://developers.shopware.com/developers-guide/rest-api/api-resource-orders/api-resource-customer`.

Correct Link would be `https://developers.shopware.com/developers-guide/rest-api/api-resource-customer/`


tiagojsag:
@kayyyy thanks for the fix

PR #112, opened 2015-08-26
semicolon is missing

kayyyy:
Because the missing semicolon you get an error if you try to open the theme settings


dnoegel:
Thank you!

PR #124, opened 2015-09-06
Remove tutorial section

dnoegel:
I suggest to remove the tutorial section again - guides like this have been added to the "plugin resources" before - if we want to have a "tutorial" section, we should still move the existing "tutorial-like" guides to there - e.g. store front components, password encoder, plugin quick start.


tiagojsag:
@Phil23 any comments on this, or can I merge it?

Phil23:
I think it's ok

PR #126, opened 2015-09-08
Added MediaService Developers Guide

janbuecker:


tiagojsag:
See also: https://github.com/shopware/devdocs/pull/129
Do NOT merge both

PR #130, opened 2015-09-10
use nodejs-legacy on ubuntu

hlb-schmidt:
users should never edit/create files in /usr/bin, and nodejs-legacy package provides node symlink. (and also depends on nodejs)


tiagojsag:
thanks for the contribution @hlb-schmidt

PR #131, opened 2015-09-14
Added .htaccess fallback solution

janbuecker:


tiagojsag:
@bcremer is this the correct location for the .zip file? This is not a plugin, so I'm not sure sure if it belongs in the exampleplugins folder

bcremer:
@tiagojsag Correct. It's not a ExamplePlugin and is it not installable via the PluginManager so the location of the ZIP-File is OK.

PR #138, opened 2015-09-22
Removed media fallback part because its part of the core now

janbuecker:


tiagojsag:
We should probably wait for the RC3 release to merge this

PR #143, opened 2015-09-28
Add bower usage guide

arnoldstoba:
@klarstil @tiagojsag Would you please review this guide really quick? :+1:


klarstil:
@arnoldstoba Thanks for the guide, the content fits the topic quite well. I noticed that you mixed the names and format them in different ways. For example you capitalize "Bower" but there a some places where you write it in lowercase. It's even worse with NPM. Please pick a style and stick to it

tiagojsag:
@arnoldstoba maybe it would be relevant to add a little example of how the users can include those JS libs in the actual template. This is already documented somewhere else in the docs, so just a link saying something like "Now that your library is installed you can add it to your template using these instructions (link)", so that the reader knows what are the next steps after this.

And regarding @klarstil comment, please check on other doc pages to see what style we use for those words, and use the same, so we have consistency across all docs

arnoldstoba:
Alright @klarstil @tiagojsag , i updated the "Bower" and "npm" references to the right case (in code fences it obviously stays small caps everywhere) and added a note that refers to the "CSS and JS file usage" quick tip. :+1:

klarstil:
Thanks @arnoldstoba for the adjustments. @tiagojsag you can merge it when you're fine with the grammar

PR #150, opened 2015-10-02
Blog post: What is an item

dnoegel:
It was a bit difficult to decide for a "canonical" term here - but I went for "item" in the blog post (only the branch name is still "product").

Basically I just wanted to show some relevant item tables, summarize the hierarchy of the new storefront "products" and talk about some other common missconceptions / questions.


bcremer:
Nice writeup :+1: 

I would like to delay publication until Wednesday 7. October because the latest blog post by Arnold was published very recently.

dnoegel:
Ah, sure; updated the date(s)

PR #156, opened 2015-10-09
Disable google analytics for dev environment

bcremer:
Ping @dnoegel

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 /composer.phar
 /app/cache/*
 /app/config/sculpin_kernel_*.yml
-/app/config/sculpin_site_*.yml
 /app/logs/*
 /output_*/
 /source/components/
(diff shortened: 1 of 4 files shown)


source/_layouts/default.twig.html
```diff
@@ -148,6 +148,7 @@ <h2 class="entry-headline">
 {% block scripts_after %}
 {% endblock %}
 
+{% if site.google_analytics_tracking_id %}
```
bcremer: Include google tagmanager in conditional so it is also hidden when in dev env.

.gitignore
```diff
@@ -3,7 +3,6 @@
 /composer.phar
 /app/cache/*
 /app/config/sculpin_kernel_*.yml
-/app/config/sculpin_site_*.yml
```
bcremer: Removed from .gitignore so future changes to `app/config/sculpin_site_prod.yml` will be shown by git.

dnoegel:
Jap, was about to do it the same way

PR #158, opened 2015-10-10
change controller registration

teiling88:
rename registerController() to registerBackendController() to prevent overwrite of Shopware_Components_Plugin_Bootstrap::registerController

use registerController() function instad of subscribeEvent()


dnoegel:
Hi,

good catch. If I am not mistaken, the optional fourth parameter with the callback method could also be removed: `onGetBackendController` does only register the View and return the controller path, that should the default callback of `registerController` also do.

teiling88:
Hey,

better catch - i fixed it :-)

PR #159, opened 2015-10-16
SW-12769 - Reindexed developers guide and fixed dead links

janbuecker:
**New**
Models `/developers-guide/plugins/models/`

I reindexed the developers guide, we should talk about this structure before merging. In addition, I fixed some dead links and dead images.


source/developers-guide/plugins/models/index.md
```diff
@@ -0,0 +1,701 @@
+---
```
tiagojsag: Most, if not all of this file, covers Doctrine basic stuff that is already covered in the Doctrine docs. Please rewrite this to only include links to the official docs + Shopware specific info

janbuecker:
Closed in favor of #172.

PR #172, opened 2015-10-22
SW-12769 - Added models guide

janbuecker:


janbuecker:
Ready for review. /cc @tiagojsag

PR #174, opened 2015-10-27
readFromDb and readFromIni doesn't work together

hlohaus:
@tiagojsag Can you fix it? :)


tiagojsag:
@hlohaus as always, an explanation of what is wrong would be very welcome ;)

PR #175, opened 2015-10-28
fix typo in js subsciption example

larsbo:


klarstil:
Hey @larsbo,

good catch, thanks for your contribution.

larsbo:
no problem!
Is there an example that demonstrates a working communication between backend controller and some frontend library? I guess the demo mentioned in this slides would be good starting point: http://klarstil.com/scd-lightweight-backend-modules/.

PR #181, opened 2015-11-06
album field-name was wrong documented

SvenHerrmann:


bcremer:
Thanks again @SvenHerrmann :+1:

PR #183, opened 2015-11-08
Add best practice blog post

dnoegel:
I would like to have this for wednesday next week if nothing is scheduled, yet? The request for some "best practice" guidelines is very common, so I decided to write this blog post.

I will proof-read it again later this day - but if you have any general remarks so far: let me know. 
@tiagojsag @bcremer


dnoegel:
Thanks @tiagojsag. Actually I used "backoffice" intentionally to get rid of our faulty backend / frontend terminology - but perhaps you are right and consistency is more relevant. :+1:

PR #187, opened 2015-11-16
SW-13107 - Added CSRF documentation

janbuecker:


tiagojsag:
@janbuecker IMO there's something missing here. You give a formal description of what an CSRF attack is, and then cover the solution, but there is no logical connection between the two. If I am new to the topic, it will be confusing for me what exactly happens in a real world CSRF attack, and how the token can prevent it. So please:
- After the formal description, please add a small, real world example of how an actual CSRF attack works.
- Once the attack vector is described, you should explain what the solution is: tokens

Once you add these two, your description of the CSRF token provider included in Shopware should come a lot more naturally to the reader

tiagojsag:
@janbuecker what's the status here?

janbuecker:
Will be continued after some other important tasks.

janbuecker:
@tiagojsag you can start reviewing it again please.

janbuecker:
@dnoegel Tiago is on holidays, maybe you can take a look at it?

tiagojsag:
I will be back on the 28th. Also, unless someone gets a Shopware shop for Christmas, I doubt that a lot of people will be reading this. Can't it wait?

source/developers-guide/index.html
```diff
@@ -45,5 +46,5 @@
     <li><a href="{{ site.url }}/developers-guide/backend-components/associations/">Backend Components - Associations</a></li>
     <li><a href="{{ site.url }}/developers-guide/backend-components/listing-extensions/">Backend Components - Listing extensions</a></li>
     <li><a href="{{ site.url }}/developers-guide/backend-components/batch-processes/">Backend Components - Batch processes</a></li>
-    <li><a href="{{ site.url }}/developers-guide/backend-extension/">Extending the backend</a></li>
```
tiagojsag: ??

janbuecker: Seems to be the result of a merge conflict. Added it back again.

tiagojsag:
https://github.com/shopware/devdocs/pull/210

PR #188, opened 2015-11-16
Added version history for articles

janbuecker:
I think that it is neccessary to provide developers with a version history of a document.

**Example: CSRF Protection**
Currently, the article describes how the protection works in general and how it has been implemented in the backend. 

What if we document the frontend implementation or do changes to the current backend implementation? How does a developer know, when and what has changed since he last visited this article?

For that reason, i've implemented a new meta tag called `history`. It's a simple key/value array whereas the key is the date of change, syntax `yyyy-mm-dd`, and the value is a short summary of the changes. So the meta tag could be used like this:

```
history:
  2015-11-16: creation
  2015-11-23: added frontend documentation
  2016-01-01: documented millenium bug
``` 

This will render a new `<h1>`  tag right below the article with a simple `<table>` which looks like this:

Date | Changes
-------|-------------
2016-01-01 | documented millenium bug
2015-11-23 | added frontend documentation
2015-11-16 | creation

### DIFF
--- a/source/_layouts/default.twig.html
+++ b/source/_layouts/default.twig.html
@@ -86,6 +86,26 @@
             {% endif %}
 
             {% block content_wrapper %}{% block content %}{% endblock %}{% endblock %}
+
+            {% if page.history is defined %}
+            <h2>Version History</h2>
+            <table>
+                <thead>
+                    <tr>
+                        <th>Date</th>
+                        <th>Changes</th>
+                    </tr>
+                </thead>
+                <tbody>
+                {% for changeDate, changeText in page.history|reverse(true) %}
+                    <tr>
+                        <td>{{ changeDate|date("Y-m-d") }}</td>
+                        <td>{{ changeText }}</td>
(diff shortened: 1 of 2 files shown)


source/_layouts/default.twig.html
```diff
@@ -86,6 +86,26 @@
             {% endif %}
 
             {% block content_wrapper %}{% block content %}{% endblock %}{% endblock %}
+
+            {% if page.history is defined %}
+            <h1>Version History</h1>
```
bcremer: Please use a H2

bcremer:
Looks good. Please document in README.md

janbuecker:
Changes have been pushed. Ready to merge.

bcremer:
@dnoegel From my point of view this is ready to merge. Might be useful in the future to generate a last-changed date for a general RSS etc.

Anything against merging?

dnoegel:
Nope, looks good for me :+1:

PR #191, opened 2015-11-23
add backend extension example

dnoegel:


dnoegel:
@bcremer: As said before: Tried to have a technical description of the system AND a small example - instead of ten pages of description for a huge example

PR #209, opened 2016-01-03
fixed semicolon

kayyyy:
If the **;** is not set, the theme configuration does not load.


bcremer:
Good Catch, thanks @kayyyy :+1:

PR #212, opened 2016-01-05
Add remove version tag to documents

tiagojsag:
This comes to me as a consequence of "Managing third-party dependencies with npm", which seems to replace "Managing third-party dependencies with bower". We already have a tag to identify in which SW version a certain feature was introduced, and it only makes sense that we flag old features as removed/deprecated.

This still needs a bit of CSS fixes. If both "from version" and "until version" are present, the title gets squashed to the left of the page. @klarstil can you take care of it, please?

### DIFF
--- a/source/_layouts/default.twig.html
+++ b/source/_layouts/default.twig.html
@@ -67,10 +67,17 @@
 
             <div class="info">
 
-                {% if page.shopware_version %}
-                <span class="version clearfix" data-shopware-version="{{ page.shopware_version }}">
-                    <span class="versionDisc">as of version</span>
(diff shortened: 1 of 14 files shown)


bcremer:
@tiagojsag The bower and npm articles are not bound to any shopware versions.
As a template developer you can still use bower to manage your own custom themes as well as template developers could have used npm to manage deps even before Shopware 5.1.2.

These tags could however be useful in the future.

tiagojsag:
@bcremer I assumed as much, that's why I didn't change the tags for those two documents. Just wanted to point out where the idea came from.

mitelg:
closing due to inactivity

PR #213, opened 2016-01-06
SW-13649 - Document csrf config options

janbuecker:


source/developers-guide/csrf-protection/index.md
```diff
@@ -93,4 +93,17 @@ class Shopware_Controllers_Backend_MyPlugin extends Shopware_Controllers_Backend
         ];
     }
 }
+```
+
+## Disable the protection
+
+In some cases you might want to disable the protection for the backend or frontend. You can achieve this by setting the following options in your `config.php`. By default, both options are enabled and set to `true`.
```
tiagojsag: By default, both options are enabled and set to `true`. => By default, both options are set to `true`.

PR #214, opened 2016-01-08
registerEventListener not a function

christianheyn:
StateManager.registerEventListener is not a function.
StateManager.registerListener works


bcremer:
Great catch, thank you @christianheyn #communityworks.

PR #217, opened 2016-01-12
fixed filepath variable

hvlmnns:


bcremer:
Good catch, thank you @hvlmnns :+1:

PR #219, opened 2016-01-14
Enterprise - Add MySQL user requirement on shop hosts

tiagojsag:
Enterprise - Add MySQL user requirement on shop hosts


tiagojsag:
@JanPietrzyk

JanPietrzyk:
:+1:

PR #225, opened 2016-01-18
Varnish config proof reading

tiagojsag:
@dnoegel


bcremer:
Thank you @tiagojsag.

PR #238, opened 2016-02-29
Fixed translations api doc

NiklasTeich:
Since Shopware 5.0.0 the translation api expects a shopId (instead of a localeId) ..


bcremer:
Thanks @NiklasTeich :+1:

PR #242, opened 2016-03-09
Fix typo in smarty media plugin example

larsbo:
replaced href attribute with src attribute to create a working image example ;-)


bcremer:
Sneaky error, thanks @larsbo :+1:

PR #252, opened 2016-03-22
Deleted doubled fragment

Asabs:


bcremer:
Nice catch, thank you @Asabs :+1:

PR #253, opened 2016-03-23
fixed minor issues in the event guide

uehler:
- fixed a syntax error in collect event example
- fixed spelling
- fixed missing module in example of the PostDispatch


bcremer:
Thanks a lot :+1:

PR #255, opened 2016-03-24
Fix wrong location for codetalks

t2oh4e:
Someone should tell @dnoegel that his talk will be in Berlin, not Hamburg ;)


bcremer:
Great catch @t2oh4e, I'll make sure @dnoegel will be in Berlin =D

PR #273, opened 2016-05-12
Update shopware5 media service index.md

a-winter:
Added explanation to get thumbnail urls from media service


bcremer:
Thank you @a-winter.

PR #277, opened 2016-05-17
Make the styling of code simpler

janbuecker:
This is just a styling update which, in my opinion, is more readable and makes an cleaner look overall. This is only related to `<code>` blocks.

I've removed the border, increased the padding and modified the expand button to fit the code styling.

### before
before

### after
after

### DIFF
--- a/source/assets/css/_imports/codeExpand.less
+++ b/source/assets/css/_imports/codeExpand.less
@@ -6,7 +6,6 @@ pre code {
 
 .code-expandButton {
     background-color: #E8E8E8;
-    border: 1px solid #d8d8d8;
     bottom: 0;
     color: #777777;
     cursor: pointer;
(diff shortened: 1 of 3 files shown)


bcremer:
I like it :+1:

dnoegel:
Don't like it :)

bcremer:
Old fart :older_man:

bcremer:
But seriously, I am open to revert that change if you really don't like it.

I like the border less look more, it's less noise in my opinion.

dnoegel:
Just checked some pages with the new styling - guess there is no serious problem with it. I like the more "obvious" code blocks with the borders. But as I don't have any taste anyway I will leave that decision to you, I guess :)

janbuecker:
Maybe because of the new colors. Some highlighting changed after the update to 0.9.3.

PR #278, opened 2016-05-17
Add address management guide

janbuecker:
Someone should proof read this.

### DIFF
--- a/source/developers-guide/address-management-guide/index.md
+++ b/source/developers-guide/address-management-guide/index.md
@@ -6,52 +6,245 @@ shopware_version: 5.2.0
 indexed: true
 tags:
   - address
-  - management
   - services
 ---
 
-The address management allows a customer to manage more than only one address which gets changed with every order. The customer is now able to create more address, e.g. for home and work, and use them later on in an order without loosing all existing address data. He can just change the reference to the default billing address, instead of changing it entirely.
+The address management allows a customer to manage more than only one address which gets changed with every order. The customer is now able to create more addresses, e.g. for home and work, and use them later on in an order without loosing all existing address data. He can just change the reference to the default billing address, instead of changing it entirely.
 
-# The address service
+<div class="toc-list"></div>
 
-The address service is used to manage all address entities in Shopware. It only works with models which makes it easy to comprehend, which properties are available and can be used. To know which properties are available, please refer to the model `\Shopware\Models\Customer\Address` in our source code.
+## Address Service
 
-## Creation
+The address service is used to manage all address entities in Shopware. It only works with models which makes it easy to comprehend, which properties are available and can be used. To know which properties are available, please refer to the model `\Shopware\Models\Customer\Address` in the source code.
 
-Like said above, you have to pass an already complete address model and an existing customer to the address service. In case of a validation error or other errors, you'll get an exception.
+Please don't handle the persisting using Doctrine ORM yourself as you might risk data inconsistency.
 
-**Example**
+### Create an address
+
+Like said above, you have to pass an already complete address model and an existing customer to the address service. In case of an error, you'll get an exception.
 
 ```php
-<?php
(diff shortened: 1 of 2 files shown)


source/developers-guide/address-management-guide/index.md
```diff
@@ -6,42 +6,112 @@ shopware_version: 5.2.0
...
-```php
-<?php
-$customer = $this->get('models')->find(Customer::class, 1);
+Like said above, you have to pass an already complete address model and an existing customer to the address service. In case of an error, you'll get an exception.
 
+```php
 $address = new Address();
 $address->fromArray($addressData);
 
+$customer = $this->get('models')->find(Customer::class, 1);
```
bcremer: Please use the fully qualified class name (not the string but the full classname) instead of `Customer::class` as it's not imported in the current scope.

source/developers-guide/address-management-guide/index.md
```diff
@@ -6,42 +6,112 @@ shopware_version: 5.2.0
 indexed: true
 tags:
   - address
-  - management
   - services
 ---
 
 The address management allows a customer to manage more than only one address which gets changed with every order. The customer is now able to create more address, e.g. for home and work, and use them later on in an order without loosing all existing address data. He can just change the reference to the default billing address, instead of changing it entirely.
```
arnoldstoba: should be: now able to create more **addresses** :)

source/developers-guide/address-management-guide/index.md
janbuecker: Does this need to be applied to the event examples too? I mind the long namespaces for TextType and so one.

janbuecker:
Updated PR with changes from @arnoldstoba and @bcremer.

source/developers-guide/address-management-guide/index.md
```diff
@@ -6,52 +6,136 @@ shopware_version: 5.2.0
...
-<?php
-$customer = $this->get('models')->find(Customer::class, 1);
+Like said above, you have to pass an already complete address model and an existing customer to the address service. In case of an error, you'll get an exception.
 
-$address = new Address();
+```php
+$address = new \Shopware\Models\Customer\Address();
 $address->fromArray($addressData);
 
+$customer = $this->get('models')->find('\Shopware\Models\Customer\Customer', 1);
```
bcremer: `\Shopware\Models\Customer\Customer` -> `\Shopware\Models\Customer\Customer::class`

janbuecker:
Updated and added example for service decoration.

source/developers-guide/address-management-guide/index.md
```diff
@@ -6,52 +6,245 @@ shopware_version: 5.2.0
...
+    }
+
+    public function setDefaultShippingAddress(Address $address)
+    {
+        $this->coreAddressService->setDefaultShippingAddress($address);
+    }
+}
+```
+
+Now that we have created the class, we need to replace the existing service with our new service. For this, subscribe to a new event in your `Bootstrap.php` file.
```
mitelg: I would not write "replace" but "decorate" 
as the service gets not really replaced. Replace and Decorate are two different things

janbuecker:
Updated PR with changes by @mitelg

source/developers-guide/address-management-guide/index.md
```diff
@@ -6,52 +6,245 @@ shopware_version: 5.2.0
...
 $address->fromArray($addressData);
 
+$customer = $this->get('models')->find(\Shopware\Models\Customer\Customer::class, 1);
+
 $this->get('shopware_account.address_service')->create($address, $customer);
 ```
 
+### Update an address
 
+Updating an address is much the same like creating one. The only difference is, that you don't have to provide the customer anymore since the address is already associated with him.
```
arnoldstoba: This should probably be rephrased too. Maybe "Updating an address is almost the same as creating one." ?

PR #286, opened 2016-05-27
Add collection of useful code snippets

FiveDigital:
As discussed in the IRC channel with shyim and mtelgmann we want to use this dev docs page to collect some useful code snippets.

### DIFF
--- a/source/developers-guide/code-snippets/index.md
+++ b/source/developers-guide/code-snippets/index.md
@@ -1,0 +1,18 @@
+---
+layout: default
+title: Code snippet collection
+github_link: developers-guide/code-snippets/index.md
+shopware_version: 5.2.0
+indexed: true
+---
+
+This article will provide some useful code snippets that can help you when developing Shopware plugins. Please feel free to add additional snippets via GitHub!
+
+## Read  config from a different shop than the current one
+
+```php
+/** @var \Shopware_Components_Config $config */
+$config = \Shopware()->Container()->get('config');
+$config->setShop($shop);
+$config->get('nameDerConfig');
+```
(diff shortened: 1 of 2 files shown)


mitelg:
@FiveDigital thanks for your PR
I think a section for each minimum SW version would be nice

source/developers-guide/code-snippets/index.md
```diff
@@ -0,0 +1,18 @@
...
+---
+
+This article will provide some useful code snippets that can help you when developing Shopware plugins. Please feel free to add additional snippets via GitHub!
+
+## Read  config from a different shop than the current one
+
+```php
+/** @var \Shopware_Components_Config $config */
+$config = \Shopware()->Container()->get('config');
+$config->setShop($shop);
```
dnoegel: I think this should be avoided in normal shopware stacks, as it will change the shop context and will affect any code running afterwards and accessing shop configurationss, even templates.

I guess as a workaround the config object could be cloned - otherwise the config reader should be refactored in a way, that it operates stateless.

FiveDigital: @mitelg Okay this example obviously sucks and I will add a new example to the PR.

@dnoegel I think there is a need for a `ConfigService` that offers a `getByContext()` method.  Or a `get(ShopContext $context = null)` method that uses the current `ShopContext` by default. If you need a real life example where we would use the new service please ask me!

dnoegel: @FiveDigital, I agree, I totally see the need for that, just wanted to prevent people from putting this as "best practice" when there are side effects.

FiveDigital:
Absolutely! I've just chosen this example because I have asked in the IRC how to get shop specific context which was the origin of the idea to collect code snippets in the devdocs. As soon as I've got an idea for a better example, this code will be code.

source/developers-guide/code-snippets/index.md
FiveDigital: Absolutely! I've just chosen this example because I have asked in the IRC how to get shop specific context which was the origin of the idea to collect code snippets in the devdocs. As soon as I've got an idea for a better example, this code will be code.

mitelg:
hey @FiveDigital 
I will close this PR as we have a new section for useful code snippets here: https://developers.shopware.com/developers-guide/cheat-sheet/
If you want to add something, feel free to open a new PR :blush: :+1: 
thank you :+1:

PR #297, opened 2016-06-09
Add plugin download for attribute system guide

OliverSkroblin:


janbuecker:
Please add `<div class="toc-list"></div>` right before `## Services`.

mitelg:
do we still have to merge this?

PR #300, opened 2016-06-16
Document CSRF plugin compatibility for older versions

janbuecker:


source/developers-guide/csrf-protection/index.md
```diff
@@ -107,3 +107,25 @@ In some cases you might want to disable the protection for the backend or fronte
 ],
 ...
 ```
+
+## Plugin compatibility for older versions
+
+Since Shopware 5.2, your plugin needs to whitelist an action in order, for example, transfer files or show a page within an window or iframe but the interface `CSRFWhitelistAware` isn't available in versions prior to 5.2 and you receive an exception. In this case, you have to create a dummy interface which will only be loaded if the it does not exists.
```
mitelg: ...within a window...
first sentence is very long.
second part of second sentence says what? :smile:

source/developers-guide/csrf-protection/index.md
```diff
@@ -107,3 +107,25 @@ In some cases you might want to disable the protection for the backend or fronte
...
+<?php
+
+namespace Shopware\Components;
+
+if (!interface_exists('\Shopware\Components\CSRFWhitelistAware')) {
+    interface CSRFWhitelistAware {}
+}
+```
+
+#### 2. Require this file above your class definition in your `Bootstrap.php`
```
mitelg: ...above in your class definition of your Bootstrap.php

janbuecker: it should not be *in* the definition - like it says: above

mitelg: ah now I see :innocent:

source/developers-guide/csrf-protection/index.md
```diff
@@ -107,3 +107,25 @@ In some cases you might want to disable the protection for the backend or fronte
 ],
 ...
 ```
+
+## Plugin compatibility for older versions
+
+Since Shopware 5.2, your plugin needs to whitelist an action in order to e.g. transfer files or show a page within an window or iframe. For Shopware versions prior to 5.2, the interface `CSRFWhitelistAware` isn't available and you'll receive an exception. In this case, you have to create a dummy interface which will only be loaded if it does not exists.
```
mitelg: In that case, you have to create a dummy interface which will only be loaded if the original one does not exists.

PR #323, opened 2016-07-15
SW-15624 / new layout structure

Phil23:


bcremer:
@Phil23 The  build is failing due to a defekt LESS, see: https://travis-ci.org/shopware/devdocs/builds/144953569#L352

```                                                       
  [Less_Exception_Parser]                              
   in basic.less on line 493, column 9                 
  491|         color: #677c8a;                         
  492|                                                 
  493|         @media screen and (min-width: 480px) {  
  494|             width: 32%;                         
  495|             margin-right: 2%;                   
  496|             margin-bottom: 0;                   
```

janbuecker:
This is a problem when using php7. Works fine with php5.6

bcremer:
Still the problem is related to this PR as the last build on travis using PHP 7 ran fine: https://travis-ci.org/shopware/devdocs#L349

janbuecker:
The php less compiler is there problem here. I've tested some other less compilers (online, node) and they all work fine with these files.

bcremer:
Ok, managed to narrow the problem down.
It's the base64 encoded background images (` background: url('data:image/png;base64,....`) that are producing the error.

PR #328, opened 2016-07-25
reference limit query parameter in API index

pajod:
Afaict the implicit limit of 1000 is not mentioned in the docs so far.
For certain purposes, it is not unreasonable to simply download all objects at once.


mitelg:
fixed with #484

PR #341, opened 2016-08-17
Add dynamic sitemap

MarcelSchmaeing:


source/sitemap.xml
```diff
@@ -0,0 +1,55 @@
...
+    <url>
+        <loc>{{ site.url }}{{ post.url }}</loc>
+        <lastmod>{{ post.date|date('c') }}</lastmod>
+        <changefreq>weekly</changefreq>
+        <priority>1.0</priority>
+    </url>
+    {% endfor %}
+
+    {% for page in data.sysadmin %}
+        {% if page.github_link %}
```
bcremer: Why do you test for `github_link` here? Wouldn't  `indexed` not be more appropriate?

bcremer:
I had to revert this patch because this broke generation of all subpages. To reproduce please remove the output_dev directory after your patch.

PR #389, opened 2016-10-14
Added frontend icons overview to designers guide

wolv-dev:
My PR was deleted, don't know why. Here again.

I now know that the overview could be found here:
https://developers.shopware.com/styletile/_components-icon-set.html
(I didn't know there was a navigation to navigate to the style components before)

But on designers guide page you'll find backend icon overview, so there should be a frontend icon overview, too. It's more intuitive this way und could be found easier on google (was looking for icon set in frontend, got no results).


Phil23:
Hey @wolv-dev,

thank you for the PR. I would prefer this solution:
https://github.com/shopware/devdocs/pull/401/files

PR #400, opened 2016-10-25
Add overlay for search result

klarstil:
The user can easily lose the focus and don't realise the search result is already rendered. Using an overlay we get a much better focus on the actual search result.

Check out the preview to get an idea how it will look like:
Preview


klarstil:
//cc @MarcelSchmaeing

PR #403, opened 2016-10-26
Fixed getting entity

nlubisch:


bcremer:
Nice catch @nlubisch :+1: 
For reference: https://github.com/shopware/shopware/blob/v5.2.9/engine/Shopware/Components/Model/EventSubscriber.php#L198

PR #406, opened 2016-10-27
Add automatic controller registration documentation

SimonBaeumer:
Shopware `> 5.2.7` automatically registers controllers, I think it should be documented :)

Further I removed invalid XML-comments which lead to copy-paste errors and added another example plugin.

### DIFF
--- a/source/developers-guide/plugin-system/index.md
+++ b/source/developers-guide/plugin-system/index.md
@@ -389,7 +389,7 @@ class SwagControllerExample extends Plugin
 ### Controller auto-registration
 
 As of version `5.2.7` there is no need for a subscriber which registers your controllers. This will be automatically done by the Shopware core.
-Just place a controller in `SwagControllerExample/Controllers/*` and you are done. 
+Just place a controller in `SwagControllerExample/Controllers/Backend|Frontend|Widgets|Api` and you are done. 
 The registration of the template would be done, i.e. in the `preDispatch()`-Method of your controller.
 
 ```php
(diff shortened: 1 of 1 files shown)


source/developers-guide/plugin-system/index.md
```diff
@@ -386,6 +386,26 @@ class SwagControllerExample extends Plugin
 }
 ```
 
+### Controller auto-registration
+
+As of version `5.2.7` there is no need for a subscriber which registers your controllers. This will be automatically done by the Shopware core.
+Just place a controller in `SwagControllerExample/Controllers/*` and you are done.
```
mitelg: you should also mention that a controller must be in a folder like "Frontend" "Backend" etc. just like in the core. otherwise the convention will not work

SimonBaeumer: Done ;)

source/developers-guide/plugin-system/index.md
```diff
@@ -386,6 +386,26 @@ class SwagControllerExample extends Plugin
 }
 ```
 
+### Controller auto-registration
+
+As of version `5.2.7` there is no need for a subscriber which registers your controllers. This will be automatically done by the Shopware core.
+Just place a controller in `SwagControllerExample/Controllers/Backend|Frontend|Widgets|Api` and you are done.
```
janbuecker: if you're using regex, use it the right way ;) `SwagControllerExample/Controllers/(Backend|Frontend|Widgets|Api)`

janbuecker: To make use of it, create a file like `SwagControllerExample/Controllers/(Backend|Frontend|Widgets|Api)/YourControllerName.php` and it follow our controller naming conventions. After that, you'll be able to call `YourControllerName`.

SimonBaeumer: done

PR #408, opened 2016-10-28
SW-16552/add api exampleplugin and guide

taltholtmann:


source/developers-guide/rest-api/plugin-api-extension/index.md
```diff
@@ -0,0 +1,557 @@
...
+
+### The File and directory structure
+<img src="img/file_structure.jpg" alt="API endpoint file structure"/>
+
+## Plugin files
+For our REST API example we only need a few files. For more information about necessary files and the 5.2 plugin system
+see the [5.2 plugin guide](/developers-guide/plugin-system).
+
+### plugin.xml
+contains the base information about the plugin, the label in englisch and german, the plugin version and the required
```
mitelg: I would leave this part out. it is not really part of the API topic

source/developers-guide/rest-api/plugin-api-extension/index.md
```diff
@@ -0,0 +1,557 @@
...
+     */
+    public function onEnlightControllerFrontStartDispatch()
+    {
+        $this->container->get('loader')->registerNamespace('Shopware\Components', $this->getPath() . '/Components/');
+    }
+}
+```
+This is our plugin bootstrap which subscribes to two events. For the one thing it uses the
+`Enlight_Controller_Dispatcher_ControllerPath_Api` event to register the API controller and for 
+the other thing it uses the `Enlight_Controller_Front_StartDispatch` event to register our resource to Shopware.
```
mitelg: I would add a reason for registering an additional namespace: we have to that, so the API Manager class could load your recourse with `\Shopware\Components\Api\Manager::getResource('Producer')`

source/developers-guide/rest-api/plugin-api-extension/index.md
```diff
@@ -0,0 +1,557 @@
...
+
+### Components/Api/Resource/Producer.php
+The resource gets called by our controller. Every controller action relies on one method of our resource. 
+* indexAction -> getList() -> returns a list of producers
```
mitelg: here you are talking about producer, but earlier in your text you call it manufacturer. you should be consistent here

exampleplugins/5.2/SwagProducerApi/Components/Api/Resource/Producer.php
```diff
@@ -0,0 +1,180 @@
...
+
+use Shopware\Components\Api\Exception as ApiException;
+use Shopware\Models\Article\Supplier as SupplierModel;
+
+/**
+ * Class Producer
+ *
+ * @package Shopware\Components\Api\Resource
+ */
+class Producer extends Resource
```
bcremer: Please choose another example. We already have a resource for the `Supplier` Model, it was named manufacturer, see here: https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Components/Api/Resource/Manufacturer.php#L46

PR #418, opened 2016-11-14
Fixed paths in controller namespaces

soebbing:
The example namespaces for controllers were singular but need to be in plural form. This PR fixes this.

Have fun
Hendrik


janbuecker:
Good catch. Thanks!

PR #419, opened 2016-11-15
Replace smarty append and prepend with smarty block parent

taltholtmann:


source/designers-guide/getting-started/index.md
```diff
@@ -128,19 +128,25 @@ You can find a complete guide to the template inheritance in the [Smarty 101 gui
...
 
-There are 3 ways to interact with blocks and add your changes inside your new template file:
+<div class="alert alert-warning">
+ In Smarty there are 3 ways to interact with blocks and add your changes inside your new template file:
 
 +   `append` (adds content __after__ the selected block)
 +   `prepend` (adds content __before__ the selected block)
-+   `overwrite` (without calling append/prepend, __overwrites__ the whole block)
-
++   `overwrite` (without calling append/prepend, __overwrites__ the whole block)
```
mitelg: please consider the PR of @3stadt here

source/designers-guide/getting-started/index.md
```diff
@@ -128,19 +128,25 @@ You can find a complete guide to the template inheritance in the [Smarty 101 gui
...
+ In Smarty there are 3 ways to interact with blocks and add your changes inside your new template file:
 
 +   `append` (adds content __after__ the selected block)
 +   `prepend` (adds content __before__ the selected block)
-+   `overwrite` (without calling append/prepend, __overwrites__ the whole block)
-
++   `overwrite` (without calling append/prepend, __overwrites__ the whole block) 
+ 
+In Shopware you should only use the __third option__ and override the parent blocks.
+If you want to `append` or `prepend` a block you should use `{$smarty.block.parent}` __as shown in the examples below__.
```
mitelg: If you want to `append` or `prepend` something to a block

janbuecker: we should not mention append/prepend at all.

klarstil: I disagree @janbuecker & @mitelg. `append` & `prepend` are part of the official Smarty implementation and therefore listed in official documentation. You can __still__ use `prepend` and `append` but the behavior is unpredictable if you mix `prepend` & `append`.

I would like to have a warning about our preferred way in one article and the rest of the articles should just use the correct way e.g. `{$smarty.block.parent}`.

We just have to highlight it otherwise we're probably confusing the customer / developer. You never know if they worked with Smarty and with the block system before.

source/designers-guide/getting-started/index.md
```diff
@@ -128,19 +128,26 @@ You can find a complete guide to the template inheritance in the [Smarty 101 gui
 ## Template blocks
 The `Bare` theme's HTML structure is wrapped in Smarty's `block` elements. These block elements group the frontend components into small, isolated blocks that can be edited/overwritten individually. To add changes to the elements of the `Bare` theme, you __can't__ simply write code inside your file. To customize a block, you have to call the correct Shopware block name.
 
-There are 3 ways to interact with blocks and add your changes inside your new template file:
+ In Smarty there are 3 ways to interact with blocks and add your changes inside your new template file:
```
mitelg: I would rewrite this whole paragraph. do not mention append and prepend. just describe what happens when you use  or don't use `{$smarty.block.parent}`

PR #433, opened 2016-12-01
Little typo fixed

mikewink:
"To enables access…" should be "To enable access…"


bcremer:
Thank you  @mikewink  :+1:

PR #434, opened 2016-12-01
sw-16272 Add payment install service to 5.2 plugin system article

taltholtmann:


source/developers-guide/plugin-system/index.md
```diff
@@ -520,7 +520,32 @@ SwagEmotion
 │   └── services.xml
 └──SwagEmotion.php
 ```
+## Add a new payment method
```
mitelg: since which shopware version is this possible? maybe it should be added

PR #479, opened 2017-01-11
misc - fix schema locations

teiling88:
fix the schema locations

### DIFF
--- a/source/developers-guide/plugin-extension-by-plugin/index.md
+++ b/source/developers-guide/plugin-extension-by-plugin/index.md
@@ -47,7 +47,7 @@ This file is equal to the old plugin.json.
 ```xml
 <?xml version="1.0" encoding="utf-8"?>
 <plugin xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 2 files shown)


bcremer:
Hi @teiling88,

I would prefer linking to the files on github as it is currently.
The relative path would force the plugin developer to use a specific directory to develop their plugins. Also note that the xsd path used in the xml files is not relevant for the actual validation. The validation always uses the xsd files provided by the shopware installation.

Another benefit of the github linked file is, that the developer can choose which target shopware version his plugin uses. So if a plugin targets for Shopware 5.2.* he uses the 5.2 branch on github, if it targets Shopware 5.3.* he uses the 5.3 branch for the xsd file.

teiling88:
Hey @bcremer,

this sounds good for me. Should we fix this in the example plugins: https://github.com/shopware/devdocs/blob/master/exampleplugins/5.2/SwagBannerApi/plugin.xml ?

bcremer:
@teiling88 Yes, please go ahead. I'll close this PR then.

PR #485, opened 2017-01-18
Update 2015-06-09-understanding-the-hook-system.md

danielwolf1:
fix event name (added missig s)


mitelg:
good catch @danielwolf1 :+1:

PR #486, opened 2017-01-19
remove obsolete deleting of cache

mitelg:
I don't think it is necessary to clear the cache on your own.
if you call `generateAttributeModels()`, the `regenerateAttributeProxies()` [(method)](https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Components/Model/ModelManager.php#L255) is also called which does the same

/cc @bcremer @wolv-dev


wolv-dev:
Sometimes it's not working and the models are not generated. That's why I added this section to the doc to workaround the issue to generate the models via CLI.

PR #491, opened 2017-01-29
Fix wrong getter method for the filter event return

aragon999:
Since `\Enlight_Event_EventArgs::get('return');` would return the wrong variable (in most cases `null`). 
See https://github.com/shopware/shopware/blob/5.2/engine/Library/Enlight/Event/EventArgs.php#L129 vs https://github.com/shopware/shopware/blob/5.2/engine/Library/Enlight/Collection/ArrayCollection.php#L85


dnoegel:
Good catch, thank you!

PR #498, opened 2017-01-30
Added cronjob example

nlubisch:


bcremer:
Thank you very much :)

PR #500, opened 2017-02-01
Explain exception differences

janbuecker:


source/developers-guide/shopware-config/index.md
```diff
@@ -44,74 +48,79 @@ As of Shopware 5.2.13 session locking is enabled by default. This prevents unsus
...
         ...
         'forceCompile' => false,
         ...
     ],
 ```
-This option controls the smarty template caching. Normally you have to clear your cache after 
-every change on the template, but if you set `forceCompile` to `true` your template will be
-compiled on every reload. This should be an essential option for every developer.
+
+This option controls the smarty template caching. Normally you have to clear your cache after every change on the template, but if you set `forceCompile` to `true` your template will be compiled on every reload. This should be an essential option for every developer. Keep in mind that it does have a great impact and loading times and should never be used in production.
```
mitelg: great impact on loading times

PR #511, opened 2017-02-16
Added missing event

shyim:


dnoegel:
Thanks @shyim. IIRC we intentionally did not document that one, as it is not compatible with multiple plugins (only one plugin "wins"). I am not saying we should not document it - but that was the intention behind it back then. What do the others think? Should we just add it for a complete overview?

Argee88:
I really think all of the available events (and possibly hooks as well) should be documented. It is just taking so much time to even figure out the names. If you managed to find the event you still have to find out what arguments/data you can retrieve from them... I'm pretty sure that I could speed up my development by some 20% if events were documented better.

mitelg:
as @dnoegel said, this event is quite dangerous, but I also agree with @Argee88 we should document all events. 
@shyim maybe you can add a hint or something like that below the event list, regarding the problem @dnoegel mentioned

t2oh4e:
@shyim Something like that: "Using this event is not recommended as it is not compatible with multiple plugins."

source/developers-guide/controller/index.md
```diff
@@ -167,6 +167,7 @@ The order of the events is the following:
 * Enlight_Controller_Action_PreDispatch
 * Enlight_Controller_Action_PreDispatch_Frontend
 * Enlight_Controller_Action_PreDispatch_Frontend_Index
+* Enlight_Controller_Action_Frontend_Index
```
mitelg: some hint would be nice. see comment

mitelg:
@shyim any update here?

mitelg:
closing due to inactivity

PR #515, opened 2017-02-23
update rest-api shop POST/PUT with correct params

lhwparis:
rest-api shops endpoint POST/PUT updated with correct minimum required params


mitelg:
thanks @lhwparis :+1:

PR #522, opened 2017-03-07
"exclude from exclude" directory var/cache/production_DATE/doctrine

waldicom:
Include directory "var/cache/production_DATE/doctrine" to allow PhpStorm to autocomplete getter and setter of attributes. E.g. $article->getAttribute()->getAttr4()
Otherwise PHPStorm is not able to autocomplete these functions


mitelg:
thanks for your contribution @waldicom :+1: 
you are right, and it's a good hint

PR #527, opened 2017-03-09
Fix dead links to article-resource Document

PhJu:


mitelg:
thanks @PhJu :+1: 
good catch!

PR #529, opened 2017-03-13
grammar (to recommend)

vintagesucks:
Used correct verb forms of _recommend_.


mitelg:
thanks @vintagesucks :+1:

PR #534, opened 2017-03-15
Typo

shyim:
The service name is small see https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Components/DependencyInjection/services.xml#L44


mitelg:
thanks @shyim :+1:

PR #538, opened 2017-03-24
Name and link to example plugin broken

jhit:
The proper name of the plugin SwagProductAssociation is SwagProductAssoc.
See: https://github.com/shopware/devdocs/tree/master/exampleplugins/legacy/Backend/SwagProductAssoc


mitelg:
thanks @jhit :+1: good catch

PR #549, opened 2017-04-05
Updated the type of the "changed" field

hkcomputer:
Changed the described type for the "changed" field from date/time to date.
It's a "date" field in the database.


mitelg:
good catch @hkcomputer :+1: Thanks!

PR #551, opened 2017-04-06
Add `active` attribute to mainDetail section

lutuh:
This attribute is needed for the article to be available. https://forum.shopware.com/discussion/comment/192651


dnoegel:
Thank you, @lutuh

PR #556, opened 2017-04-25
Fetching plugin-path from service parameter

oktupol:
I think using the container parameter is better than getting the plugin and invoking getPath on it. What do you think?


mitelg:
hi @oktupol 
thanks for your PR :+1: 
you are right. this is a better way!

PR #575, opened 2017-05-16
Add documentation about the CSRF ajax option

jenskueper:


source/developers-guide/csrf-protection/index.md
```diff
@@ -137,3 +137,35 @@ if (!interface_exists('\Shopware\Components\CSRFWhitelistAware')) {
 ```
 require_once __DIR__ . '/Components/CSRFWhitelistAware.php';
 ```
+
+## Protect GET requests via CSRF tokens
+
+Since Shopware 5.2.22 you are able to protect your AJAX JSONP GET requests via CSRF tokens. There are two main steps to enable CSRF support for your Shopware Plugin. 
+
+The first step is to implement the `CSRFGetProtectionAware` interface for the controller which contains the action you want to protect via CSRF. The interface requires a method called `getCSRFProtectedActions` which returns all actions that will be validated. Your controller should now look like this:
```
janbuecker: we should document the forward-compatibility by shipping the `CSRFGetProtectionAware` interface with the plugin like in the section above.

PR #578, opened 2017-06-03
Fix typo in PUT method

dciccale:
PUT request section should describe orders endpoint and parameters.


janbuecker:
Absolutely correct. Thanks @dciccale 👍

PR #587, opened 2017-06-13
styling changes.

martijnsnijder1:
- cab -> may be
- altered some "can" to "may"


janbuecker:
Thanks @martijnsnijder1

PR #593, opened 2017-06-22
Changed docs

gusdeboer:
"die('Hello World')" single quotes are used in the example above.

The description speaks of more then one request so the correct form is "these"


mitelg:
thanks @gusdeboer :+1:

PR #599, opened 2017-06-29
Missing Field for posting a new address

ddoehring:
You need to specifiy the missing field customer to post a new address.
-- Added customer (integer -> foreign key)


mitelg:
thanks @ddoehring :+1:

PR #611, opened 2017-07-12
fixed broken link "5.3 signature"

vintagesucks:
previous url led to a 404 page


janbuecker:
The link is just missing a leading forward slash. We do not add the full domain here as it will be added when the docs are getting generated. The URL is dynamic.

vintagesucks:
You're right – fixed that. 👍

mitelg:
thanks @vintagesucks :+1:

PR #622, opened 2017-07-24
5.2 plugin system: added frontend resources auto-registration

buddhaCode:
see https://github.com/shopware/shopware/blob/5.4/UPGRADE-5.2.md#5213


mitelg:
thanks for your PR @buddhaCode :+1:

PR #623, opened 2017-07-25
Improve config settings of elasticsearch

aherking:
when you copy&paste it to your existing config.php, the first array cling square causes problems and the settings will not be used. the view should be better with an example config.php


mitelg:
thanks @aherking :+1:

PR #624, opened 2017-07-25
fix typo on shopware 5 cli example

buddhaCode:


source/developers-guide/shopware-5-cli-commands/index.md
```diff
@@ -48,8 +48,7 @@ class Shopware_Plugins_Frontend_SwagExample_Bootstrap extends Shopware_Component
     public function install()
     {
         $this->subscribeEvent(
-            'Shopware_Console_Add_Command',
-            'onAddConsoleCommand'
+            'Shopware_Console_Add_Command' => 'onAddConsoleCommand'
```
mitelg: this is not the correct syntax for this method. see here: https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Components/Plugin/Bootstrap.php#L434

As this is the old plugin system, the example is correct

buddhaCode: gosh. ur right.

PR #632, opened 2017-08-08
Add required plugins documentation

mitelg:


source/developers-guide/plugin-system/index.md
```diff
@@ -619,6 +619,23 @@ Example `plugin.xml`:
 </plugin>
 ```
 
+#### Require other plugins
+If you need other plugins to be installed, then you are able to define them in your plugin.xml file like this
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<plugin xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/shopware/5.2/engine/Shopware/Components/Plugin/schema/plugin.xsd">
```
janbuecker: 5.3 pls :)

PR #642, opened 2017-08-22
Added missing model for Order resource (Payment)

ddoehring:
When getting a specific Order, you also get a Payment-model.
This model was missing, so I added it ;).

### DIFF
--- a/source/developers-guide/rest-api/models/index.md
+++ b/source/developers-guide/rest-api/models/index.md
@@ -564,8 +564,8 @@ subgroup: REST API
 
 ## Payment
 
-* **Model:** Shopware\Models\Customer\Payment
-* **Table:** s_core_payment
+* **Model:** Shopware\Models\Payment\Payment
+* **Table:** s_core_paymentmeans
 
 ### Structure
 
--- a/source/developers-guide/rest-api/models/index.md
+++ b/source/developers-guide/rest-api/models/index.md
@@ -562,6 +562,30 @@ subgroup: REST API
 | group                    | string                  |                                                 |
 | sendMail              | boolean                  |                                                 |
 
+## Payment
+
+* **Model:** Shopware\Models\Customer\Payment
+* **Table:** s_core_payment
+
+### Structure
+
+| Field               | Type                  | Original object                                 |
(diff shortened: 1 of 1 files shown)


source/developers-guide/rest-api/models/index.md
```diff
@@ -562,6 +562,30 @@ subgroup: REST API
 | group                    | string                  |                                                 |
 | sendMail              | boolean                  |                                                 |
 
+## Payment
+
+* **Model:** Shopware\Models\Customer\Payment
```
mitelg: this model and this table don't exist in Shopware. therefore I will decline your PR. 

I'm not quite sure, what your intention was behind that, but feel free to open another PR :blush: :+1:

ddoehring: `
### {
    "id":"",
    "name":"",
    "description":"",
    "template":"",
    "hide":"",
    "additionalDescription":"",
    "debitPercent":"",
    "surcharge":"",
    "surchargeString":"",
    "position":"",
    "active":"",
    "esdActive":"",
    "mobileInactive":"",
    "pluginId":""
}
--

So this is the JSon structure I'm getting back from calling http://sw-demo.dev.1drop.de/api/orders/2 (i left out the values in the response).

I didn't find a model, that matches this structure but in the docs, Payment is mentioned and linked to (at least it tries). So I thought it's missing, isn't it :smile:?

`

mitelg: this looks like `\Shopware\Models\Payment\Payment` and the table `s_core_paymentmeans`
It's missing, you are right. 

indeed the link on this page https://developers.shopware.com/developers-guide/rest-api/api-resource-orders/ seems to be wrong. `Payment` redirects to `Payment-Instance` which is wrong.

Maybe you can change these both things? then we have the correct state :blush: :+1:

ddoehring:
Sadly i was unable to extends this PR to include my second PR (fixing payment model link) 😃. So i made a new PR.
Both Pull Requests have no changes, which all together fixes the mentioned issues.

Sorry for this inconvinience 😃

mitelg:
thanks @ddoehring :+1:

PR #643, opened 2017-08-22
Added missing Get (List)-operation for orders

ddoehring:
Get (List)-operation for orders was missing --> Added.

### DIFF
--- a/source/developers-guide/rest-api/api-resource-orders/index.md
+++ b/source/developers-guide/rest-api/api-resource-orders/index.md
@@ -39,7 +39,7 @@ This API call requires one of the following parameters to be defined:
 
 | Model                              | Table                 |
 |------------------------------------|-----------------------|
-| Shopware\Models\Customer\Customer  | s_user                |
+| Shopware\Models\Order\Order        | s_order               |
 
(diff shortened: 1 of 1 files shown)


source/developers-guide/rest-api/api-resource-orders/index.md
```diff
@@ -86,6 +86,54 @@ This API call requires one of the following parameters to be defined:
...
+#### Required Parameters
+
+For this operation, no parameters are required. To get a list of all orders, simply query:
+* **http://my-shop-url/api/orders**
+
+### Return Value
+
+| Model                              | Table                 |
+|------------------------------------|-----------------------|
+| Customer  | s_user                |
```
soebbing: The Order-endpoint returns "Order" objects coming from the s_order-table. And yes, the documentation for retrieving a single order is wrong as well. So if you want to extend your PR to include a fix for that error, it would be very much appreciated! 😄

ddoehring:
I have extended my previous PR ;)

soebbing:
Thank you for your help, @ddoehring and making Shopware (and it's documentation) better! ❤️

PR #652, opened 2017-09-05
[REST-API] Fixed GET schema for category

ddoehring:
Fixed schema for GET operation. The parameter 'external' is a string instead of a boolean.


mitelg:
thanks for your contribution @ddoehring :+1:

PR #655, opened 2017-09-06
[REST-API] Fixed a little optical issue

ddoehring:
The appearance of the table in the order's model was screwed up.
The problem was only visible when switching from soft-wrap to no-wrap.


mitelg:
thanks @ddoehring :+1:

PR #656, opened 2017-09-07
[REST-API] Fixed field for Billing model

ddoehring:
Billing does require a country object instead of just a countryId.


ddoehring:
Never mind, i was wrong. API is correct, I was wrong 😄

PR #672, opened 2017-10-04
Fix memcache/memcached Naming

Isengo1989:
Fix memcache/memcached Naming (since you explain how to use and install memcached, not memcache)


mitelg:
thanks @Isengo1989 :+1: 
good catch

PR #675, opened 2017-10-23
Add unknown {s} tag documentation

OliverSkroblin:


janbuecker:
The format of the list looks a bit broken. Please fix that issue.

MarcelSchmaeing:
can you add a hint that this can be configured in the config.php also refer to the config.php documentation.

PR #678, opened 2017-10-24
Add "Community provided Libraries" for Smarty

screeny05:
i don't believe that there are more of this kind


screeny05:
thanks for your clear stance on this!

PR #683, opened 2017-10-26
Update labs architecture document

janbuecker:
Updated architecture page with latest changes


source/labs/housekeeping/index.md
```diff
@@ -25,3 +26,231 @@ We've restored out testing environment to make sure that ported components still
...
+
+As already mentioned, the minimum required version is MySQL 5.7 and we still don't support other databases like PostgreSQL, MSSQL, ...
+
+### Conventions
+
+We've added simple conventions to keep everything straight from the beginning.
+
+- The `s_` prefix is gone
+- Every table must be singular for problems like `s_articles_details`
+- Every table and column must be lower-snake-case
```
bcremer: "lower-snake-case" is actually called [kebab-case](https://en.wikipedia.org/wiki/Letter_case#Special_case_styles) 🍌

janbuecker: image
nooooope, because it still has underscores instead of hyphens :p

bcremer: Then your example string should be "lower_snake_case"....

PR #693, opened 2017-11-09
change doku to plugin 5.2 system / backend components details

swDennis:


exampleplugins/5.2/SwagProductDetail/SwagProductDetail.php
```diff
@@ -0,0 +1,79 @@
...
+    public function uninstall(UninstallContext $uninstallContext)
+    {
+        if (!$uninstallContext->keepUserData()) {
+            $this->removeDatabase();
+        }
+    }
+
+    private function createDatabase()
+    {
+        $modelManager = $this->container->get('models');
```
mitelg: maybe you could extract this into a helper method, because you do it twice

PR #695, opened 2017-11-10
Migrate to 5.2 plugin system: listing extension

swDennis:


exampleplugins/5.2/SwagProductAssoc/SwagProductAssoc.php
```diff
@@ -47,11 +48,7 @@ private function createDatabase()
         $modelManager = $this->container->get('models');
```
mitelg: this should also be part of the helper method. change everywhere :wink:

PR #697, opened 2017-11-13
Migrate the doku / extend customer module

swDennis:


exampleplugins/5.2/SwagExtendCustomer/Resources/views/backend/swag_extend_customer/view/detail/window.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+});
+//{/block}
+
+
+
+
+
+
```
mitelg: much empty lines. :dog:

exampleplugins/5.2/SwagExtendCustomer/Resources/views/backend/swag_extend_customer/view/detail/window.js
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Replace the textBox for field "title" with a comboBox
+     *
+     * @returns { Ext.form.FieldSet }
+     */
+    createPersonalFieldSet: function () {
+        var me = this,
+            fieldSet = me.callParent(arguments);
+
+        fieldSet.items.items[1].items.items[0] = Ext.create('Ext.form.field.ComboBox', {
```
mitelg: do you overwrite the existing salutation field? maybe we should identify it in another way, so it is safe if other plugins do a similar extension

PR #700, opened 2017-11-13
rename additionnaltext to additionaltext

scripter40:


mitelg:
thanks @scripter40 :+1: :tada:

PR #704, opened 2017-11-14
Migrate doku - Emotion elements

swDennis:


source/developers-guide/custom-shopping-world-elements/index.md
```diff
@@ -20,18 +20,24 @@ The shopping worlds are one highlight feature of Shopware. Here you can edit var
...
+        'name' => 'Vimeo Video',
+        'xtype' => 'emotion-components-vimeo',
+        'template' => 'emotion_vimeo',
+        'cls' => 'emotion-vimeo-element',
+        'description' => 'A simple vimeo video element for the shopping worlds.'
+    ]
+);
 ```
-In the `install()` method of our plugin we register a new element and save it in the variable `$vimeoElement` for later use. The `createEmotionComponent()` method expects a configuration array with the following properties:
+In the `install()` method of our plugin we register a new element. The `createOrUpdate()` method expects the plugin name, a element name and a configuration array with the following properties:
```
mitelg: you use a helper class in the install method which creates the new element

source/developers-guide/custom-shopping-world-elements/index.md
```diff
@@ -463,32 +465,61 @@ public function handle(ResolvedDataCollection $collection, Element $element, Sho
 
 Keep in mind to use a unique key for requesting and getting products. For best practise, use the element's id in your key (`$element->getId()`).
 
-### Shopware 5.2 and below
+### Shopware 5.2
```
mitelg: is it still necessary to point out that this is for 5.2?

source/developers-guide/custom-shopping-world-elements/index.md
swDennis: D

PR #716, opened 2017-11-16
Migrate core service extension documentation

philipgatzka:
Adjusted documentation for core service extension via plugin to fit the new plugin system.

### DIFF
--- a/source/developers-guide/shopware-5-core-service-extensions/index.md
+++ b/source/developers-guide/shopware-5-core-service-extensions/index.md
@@ -55,7 +55,7 @@ In most cases, however, custom services want to extend the default behaviour, ra
 </container>
 ```
 
-The plugins __SwagExample\Bundle\StoreFrontBundle\ListProductService__ gets the existing implementation as a constructor argument, and can use it internally in its own logic. Another great advantage of using a decorator pattern over a full service replacement is that multiple implementations can provide their only logic on top of the previously existing service, regardless of it being the core service itself or an already decorated version/replacement of it.
+The plugins __SwagExample\Bundle\StoreFrontBundle\ListProductService__ gets the existing implementation as a constructor argument, and can use it internally in its own logic. Another great advantage of using a decorator pattern over a full service replacement is that multiple implementations can provide their own logic on top of the previously existing service, regardless of it being the core service itself or an already decorated version/replacement of it.
 
 In the next paragraphs we will further explain how and why you should implement each of these approaches, and provide demo code that you can use as a base for your own implementation.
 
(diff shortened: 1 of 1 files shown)


source/developers-guide/shopware-5-core-service-extensions/index.md
```diff
@@ -16,108 +16,125 @@ menu_order: 130
...
-}
+```xml
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="shopware_storefront.list_product_service" class="SwagExample\Bundle\StoreFrontBundle\ListProductService" />
```
mitelg: did you tested this? I'm not sure, but afaik you also have to use the decorate functionality to replace a service

source/developers-guide/shopware-5-core-service-extensions/index.md
```diff
@@ -16,108 +16,125 @@ menu_order: 130
...
+                 public="false">
+
+            <argument type="service" id="swag_example.list_product_service.inner"/>
+        </service>
+    </services>
+</container>
 ```
 
-The __DecoratorServiceImplementation__ gets the existing implementation as a constructor argument, and can use it internally in its own logic. Another great advantage of using a decorator pattern over a full service replacement is that multiple implementations can provide their only logic on top of the previously existing service, regardless of it being the core service itself or an already decorated version/replacement of it.
+The plugins __SwagExample\Bundle\StoreFrontBundle\ListProductService__ gets the existing implementation as a constructor argument, and can use it internally in its own logic. Another great advantage of using a decorator pattern over a full service replacement is that multiple implementations can provide their only logic on top of the previously existing service, regardless of it being the core service itself or an already decorated version/replacement of it.
```
mitelg: ... can provide their own logic on top ...

source/developers-guide/shopware-5-core-service-extensions/index.md
```diff
@@ -350,101 +335,84 @@ However, you can use service decoration for tasks other than this.
...
 
-This doesn't mean you need to always decorate existing services. In some scenarios, full replacement might be a better solution than decorating. Just keep in mind that other plugins might want to decorate your service as if it where the core implementation, to ensure both your and other plugins work as expected.
+This doesn't mean you always need to decorate existing services. In some scenarios, full replacement might be a better solution than decorating. Just keep in mind that other plugins might want to decorate your service as if it were the core implementation, to ensure both your and other plugins work as expected.
```
mitelg: ... as if your service is the core implementation ....

PR #722, opened 2017-11-17
[FEATURE] Changed example and added instructions for better understan…

wolv-dev:
…ding


mitelg:
thanks  @wolv-dev :+1: 
it could indeed be a little bit confusing :wink:

PR #725, opened 2017-11-20
Migrate doku - Extend plugin by plugin

swDennis:


source/developers-guide/plugin-extension-by-plugin/index.md
```diff
@@ -83,90 +83,46 @@ Make sure that Subscriber-Services implement the tag.
...
+<service id="swag_extend_custom_products.file_upload.file_type_whitelist_decorator"
+         class="SwagExtendCustomProducts\Decorators\FileTypeWhiteListDecorator"
+         decorates="custom_products.file_upload.file_type_whitelist"
+         public="false">
+    <argument type="service" id="swag_extend_custom_products.file_upload.file_type_whitelist_decorator.inner"/>
+</service>
 ```
-If Custom Products (v2) calls the event, set the decorator into the container to replace the original service.
+
+The container replaced the original service with our. If the service is called the  'FileTypeWhiteListDecorator' is returned.
```
mitelg: ... replaces the ... with our own.

mitelg:
in the plugin.xml it is possible to define the required plugin like this:
```
    <requiredPlugins>
        <requiredPlugin pluginName="SwagCustomProducts" minVersion="4.0.0"/>
    </requiredPlugins>
```
the minVersion attribute makes sure, that we use the correct version of CustomProduct. would be nice, if you could write it in the text and in the plugin.xml :blush: :+1:

PR #727, opened 2017-11-20
Migrate risk rule example plugin

philipgatzka:


exampleplugins/5.2/SwagCustomRiskRule/SwagCustomRiskRule.php
```diff
@@ -0,0 +1,15 @@
...
+namespace SwagCustomRiskRule;
+
+use Shopware\Components\Plugin;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class SwagCustomRiskRule extends Plugin
+{
+    public function build(ContainerBuilder $container)
+    {
+        $container->setParameter('swag_custom_risk_rule.plugin_dir', $this->getPath());
```
mitelg: is not necessary, see here: https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Components/Plugin.php#L154

PR #728, opened 2017-11-20
Migrate doku - License check

swDennis:


source/developers-guide/plugin-license/index.md
```diff
@@ -36,7 +36,7 @@ public function checkLicense($throwException = true)
 }
 ```
 
-This method validates if the current shop has a valid license for your plugin. It should be placed inside your plugin's Bootstrap class, and used whenever you want to ensure that the current shop has a valid license for your plugin (for example, during plugin installation).
+This method validates if the current shop has a valid license for your plugin. It should be placed inside your plugin base class or in a service for used whenever you want to ensure that the current shop has a valid license for your plugin (for example, during plugin installation).
```
mitelg: It should be placed inside your plugin base class, services, subscribers, or whenever you want to ensure that the current shop has a valid license

swDennis: D

PR #731, opened 2017-11-20
 Article images API complement

Isengo1989:
Adding info to article/media API for easier understanding that file is an option to use images on the server


mitelg:
thanks for your contribution @Isengo1989 :+1: :tada:

PR #754, opened 2017-12-06
fixed vimeo element example

uehler:
the registration of the emotion component view subscriber was missing and a duplicate of the subscriber for the backend extension was mixed in between the description of the emotion handler

### DIFF
--- a/source/developers-guide/custom-shopping-world-elements/index.md
+++ b/source/developers-guide/custom-shopping-world-elements/index.md
@@ -30,7 +30,6 @@ $vimeoElement = $this->emotionComponentInstaller->createOrUpdate(
     'SwagVimeoElement',
     [
         'name' => 'Vimeo Video',
-        'xtype' => 'emotion-components-vimeo',
         'template' => 'emotion_vimeo',
         'cls' => 'emotion-vimeo-element',
         'description' => 'A simple vimeo video element for the shopping worlds.'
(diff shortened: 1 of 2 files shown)


source/developers-guide/custom-shopping-world-elements/index.md
```diff
@@ -419,7 +427,7 @@ class VimeoComponentHandler implements ComponentHandlerInterface
 {
     public function supports(Element $element)
     {
-        return $element->getComponent()->getType() === 'emotion-components-vimeo';
+        return $element->getComponent()->getTemplate() === 'emotion_vimeo';
     }
```
swDennis: Please revert this line. The rest is great!

uehler: But then the code won't work at that point:
I removed the xtype in the createOrUpdate()-method because you don't need it to create the element or for the handler but I mention it when it is needed to create the custom view in the "Adding a custom emotion component in ExtJS" section.
This advanced section is talked about after the handler section, so I changed from type check to template check. You don't have a custom xtype there to check for.

And if you define the xtype in the element at the beginning of the doc you can't say "[...] we already see a full functional shopping world element [...]" in the "Creating a frontend template for the element" section because the element won't work. There is no js for that xtype so the element does not open a new window with the elements settings.

uehler: @swDennis

mitelg:
@swDennis I tested the changes and it's like @uehler said the x-type is not mandatory and it works without that. 
@uehler thanks for improving our docs :+1: :tada:

PR #758, opened 2017-12-12
Deliver more info about applyDataAttributes()

buddhaCode:
Added attributes and example for applyDataAttributes() method.

### DIFF
--- a/source/designers-guide/javascript-statemanager-and-pluginbase/index.md
+++ b/source/designers-guide/javascript-statemanager-and-pluginbase/index.md
@@ -188,7 +188,7 @@ $.plugin('example', {
     * Setter method which overrides the value of the provided key with the provided value.
 * ```applyDataAttributes()```
     * **Arguments**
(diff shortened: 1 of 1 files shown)


source/designers-guide/javascript-statemanager-and-pluginbase/index.md
```diff
@@ -187,7 +187,9 @@ $.plugin('example', {
         * ```value : Mixed``` - Value for the provided key
     * Setter method which overrides the value of the provided key with the provided value.
 * ```applyDataAttributes()```
-    * Fetches the provided configuration keys and overrides the values based on the elements ```data``` attributes.
+    * **Arguments**
+        * ```[shouldDeserialize : Boolean = false]``` - Tries to parse the given string values and returns the right value if its successful. Supports boolean, null, number, json, string. (optional, default: false)
```
mitelg: the default value is not false. see here: https://github.com/shopware/shopware/blob/5.3/themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js#L311
everything else than false will result into deserializing the values

PR #763, opened 2017-12-18
Fix directory structure overview

lkreimann:
one duplicate folder in the directory structure. was "documents", should be "widgets"


mitelg:
thanks @lkreimann good catch :+1: :tada:

PR #765, opened 2017-12-21
Few changes in Models documentation

oktupol:
Most apparent change is that I replaced the Type annotations for the entity-manager from `Doctrine\ORM\EntityManager` to `Shopware\Components\Model\ModelManager` after noticing, that I am working with a `ModelManager` when I wrote an event listener. While `EntityManager` is technically correct and sufficient in most cases, I imagine it's still good to know that there is more than that.

Additionally, there are a couple of minor changes regarding code-style.

Merry Christmas 🎄  :skier:


mitelg:
thanks @oktupol :+1: :tada: 

we wish you a merry christmas, too! :snowflake:

PR #772, opened 2018-01-08
NTR - Add attribute naming notices to cheat sheet

taltholtmann:


source/developers-guide/cheat-sheet/index.md
```diff
@@ -90,6 +90,15 @@ $attributeCrudService->update(
     
 );
 ```
+### Naming - Attribute generation
+If you get the `Error: Unrecognized field: my_field` when querying for your attribute it could be that you forgot to generate the attribute models:
+```php
+Shopware()->Container()->get('models')->generateAttributeModels();
```
mitelg: please use `$this->container` instead of the singleton

PR #778, opened 2018-01-16
Fix-missing-redisAuth-parameter

mnaczenski:
Since SW-20299 it is possible to use a redis-auth, but if this parameter isn't set, the implementation won't work.


mitelg:
thanks @mnaczenski

PR #779, opened 2018-01-17
less -> fewer

oktupol:
Ser Davos would be proud


mitelg:
:D thanks @oktupol :+1: :tada:

PR #791, opened 2018-01-23
fix syntax

buddhaCode:
Attribute names will be parsed vom snake case (html) to camel case (js).

### DIFF
--- a/source/developers-guide/shopware-5-plugin-update-guide/index.md
+++ b/source/developers-guide/shopware-5-plugin-update-guide/index.md
@@ -436,7 +436,7 @@ $.plugin('yourName', {
    init: function() {
        // ...initialization code
 
-       //The applyDataAttributes function merges data attributes into plugin options. Example element: <div class="test" data-exampleValue="value2"></div>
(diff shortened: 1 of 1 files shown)


mitelg:
hey @buddhaCode 
snake case is with _ underscore. 
Afaik `example-value` and `exampleValue` are two different things. Although if you write `data-exampleValue` in your smarty code, the output will be `data-examplevalue`
So with your change, the example won't work, imho.

buddhaCode:
I mean dash case. Sorry.

And yes, your right. I proposed a pull request here: https://github.com/shopware/shopware/pull/1477
I just referred to this PR on the docs: https://github.com/shopware/devdocs/pull/758
The example in this PR doesn't work either.

mitelg:
ok, I think we have to talk about that thing internally (looking at you @klarstil :wink: )

If your PR gets merged, we should adjust the docs after that. I would like to close your PR, as your change would not work as intended. Is this ok for you?

buddhaCode:
It's fine. :)

PR #798, opened 2018-01-29
add sample for clickable label

tinect:


tinect:
So true... Now it should be correct. Sorry for that

mitelg:
thanks @tinect :+1: :tada:

PR #808, opened 2018-02-02
checkbox and boolean

Dwza:
The documentation is missing informations about type boolean and some optional tags like <option> that can be set to define e.g. checkbox and give them a predefinition.


Dwza:
Hope I stated everything right. But this is what I found out so far.

mitelg:
hey @Dwza 
thanks for your PR. As in https://github.com/shopware/shopware/pull/1512#issuecomment-363730444 stated, we do not want to support the x-type tag officially. 
Nevertheless the boolean type is still missing, so please remove the other parts and the PR will be merged

source/developers-guide/plugin-configuration/index.md
```diff
@@ -173,6 +173,45 @@ Below all supported configuration elements including their design and source cod
 
 <img src="img/numberfield.png" alt="Numberfield" class="image-border" />
 
+### Boolean
+To define something similar like checkbox you can use `boolean`
+```xml
+<elements>
+    <element required="false" type="boolean" scope="shop">
```
mitelg: just state only the required things here. label, name, etc are not special for the boolean type

mitelg: It would also nice, if you could provide an image of the boolean type

Dwza: Of course they may not be required but for more transparency when viewing the code and not start trying to guess what could be used, I really would recommend to keep these lines. I could add a sentence to point out that they are writen for transparency and not because of requirement.

source/developers-guide/plugin-configuration/index.md
```diff
@@ -173,6 +173,51 @@ Below all supported configuration elements including their design and source cod
...
+    </element>
+</elements>
+```
+Notice: Fields like `name`, `label`,`description` are not required. They just listed for code transparency.
+
+This will provide you with a prefilled dropdown.
+
+<img src="img/boolean.png" alt="Booleanfield" class="image-border" />
+
+Currently there is no element for checkbox at the moment.
```
mitelg: please remove this part as stated here: https://github.com/shopware/devdocs/pull/808#issuecomment-363733637

mitelg:
thanks for the changes :+1: 

but one last thing :wink:

PR #813, opened 2018-02-07
Fixed debug usage

shyim:
The fourth argument must be false, so dump returns the result


mitelg:
thanks @shyim :+1: :tada:

PR #816, opened 2018-02-08
fix readme notice in searchbundle

uehler:
There is no README.md in Shopware\Bundle\StoreFrontBundle, I think it should be Shopware\Bundle\SearchBundle\README.md


mitelg:
thanks @uehler you are right :+1: :tada:

PR #818, opened 2018-02-13
Remove checkout SEO description

soebbing:
The SEO routes got scraped, so we need to remove them from the upgrade guide


oktupol:
Just curious; why did you decide to remove them?

soebbing:
We explained it a bit more in the - now empty - migration: https://github.com/shopware/shopware/blob/5.4/_sql/migrations/1208-add-checkout-seo-routes.php

Basically: If you create a language subshop, it also inherits these routes and you would have to modify them by hand.

PR #827, opened 2018-02-26
Fix wording on frontpage of the styletile

Lucsy3012:
Fixed a simple wording error


mitelg:
thanks @Lucsy3012 :+1:

PR #830, opened 2018-02-27
Fix another typo

t2oh4e:
Seems like I added a typo while fixing typos. Nice one :D


mitelg:
thanks again @t2oh4e :smile:

PR #837, opened 2018-03-15
Changed array notation

gusdeboer:
The examples above use "[]" instead of "array()", should be consistent.


mitelg:
thanks @gusdeboer :+1:

PR #842, opened 2018-03-21
Correct typo: snippet namespace missed one character

jkrzefski:
As seen in the [snippet manager code](https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Components/Snippet/Manager.php#L310) the actual namespace starts with `themes/` instead of `theme/`.


mitelg:
good catch @jkrzefski :+1: :tada: thanks :blush:

PR #845, opened 2018-03-22
Fixed style tile link

oktupol:
Links to style tile now lead to `/styletile/` instead of `http://styletile.dev.shopware.in/`


mitelg:
thanks @oktupol :+1: :tada:

PR #853, opened 2018-04-13
Add a link to cloud.elastic.co

dadoonet:
Awesome integration. Nice to see it.

I'm proposing this change to make sure that installing elasticsearch is not a blocker for a production deployment in term of investment.
I hope it helps.


mitelg:
hey @dadoonet 
thanks for your first PR on this project :tada: 
we decided to close this PR, because it has nothing to do with Shopware directly and we don't want to document how to setup elastic search :wink:

dadoonet:
Fair enough.

> we don't want to document how to setup elastic search

That was exactly my point. Giving an option to your users so they don't event need to think about it.

Anyway, thank you for reading it.

PR #855, opened 2018-04-25
Correct path to risks.js

Bugius:


mitelg:
hey @Bugius 
thanks for pointing this out. the problem is that the directory name was wrong, not the loading of the template. we cannot use the same directory path in our plugin as the backend file, we want to extend.

PR #858, opened 2018-05-14
describe cache purging in cluster env

buddhaCode:
Describe settings to set, when using build-in http cache in a cluster environment. This is needed for correct cache invalidation. I have an article on my mind, were different varnish and shopware http cache scenarios are described. I would link that in the section but i couldn't find it any more. Do anybody remember this article, too?


buddhaCode:
@mitelg Is something wrong with my proposal?

mitelg:
sorry @buddhaCode totally forgot about that one :see_no_evil: 
good work :+1: :tada: :blue_heart:

PR #859, opened 2018-05-15
Correct controller name

Bugius:


mitelg:
thanks @Bugius good catch :+1: :tada:

PR #867, opened 2018-06-12
Change field name for "sortmode"

vanWittlaer:
to all small letters

### DIFF
--- a/source/developers-guide/rest-api/api-resource-property-group/index.md
+++ b/source/developers-guide/rest-api/api-resource-property-group/index.md
@@ -39,6 +39,8 @@ You can retrieve a property group by using its id
 | options              | array                  | **[Option](../models/#property-group-option)**            |
 | attribute              | array                  | **[Attribute](../models/#property-group-attribute)**        |
 
+Note the camel case notation for 'sortMode'.
+
 ## GET (List)
 
 To get more than one property group at once, simply remove the id parameter from the request URL.
@@ -66,7 +68,9 @@ You can post or put data by querying the following URL:
 | name                  | string                  |                                                            |
 | position              | integer                  |                                                            |
 | comparable          | boolean                  |                                                            |
-| sortMode              | integer                  |                                                            |
+| sortmode              | integer                  |                                                            |
+
+Note the small case notation for 'sortmode' to go with POST or PUT as opposed for GET.
 
 ## DELETE
 To delete a property group, simply query this URL with a `DELETE` request:


mitelg:
hey @vanWittlaer 
thanks for your PR. 
the fieldnames are referring to the property names of the doctrine model -> https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Models/Property/Group.php#L82
not to the actual database column name. 
So the documentation is correct here :blush: 
```
Example response
{
	"data":[
		{
			"id":1,
			"name":"Edelbr\u00e4nde",
			"position":0,
			"comparable":true,
			"sortMode":2,
			"options":[
				{
					"id":1,
					"name":"Alkoholgehalt",
					"filterable":true
				},
				{
					"id":2,
					"name":"Flaschengr\u00f6\u00dfe",
					"filterable":true
				},
				{
					"id":4,
					"name":"Farbe",
					"filterable":true
				},
				{
					"id":6,
					"name":"Geschmack",
					"filterable":true
				},
				{
					"id":7,
					"name":"Trinktemperatur",
					"filterable":true
				}
			],
			"attribute":null
		}
	],
	"total":1,
	"success":true
}
```

vanWittlaer:
Strange - how comes when using the API internally via 

`\Shopware\Components\Api\Manager::getResource('PropertyGroup')`

it requires the field name to be "sortmode"?

mitelg:
This is the result, if I use the API resource internally
```
Array
(
    [data] => Array
        (
            [0] => Array
                (
                    [id] => 1
                    [name] => Edelbrände
                    [position] => 0
                    [comparable] => 1
                    [sortMode] => 2
                    [options] => Array
                        (
                            [0] => Array
                                (
                                    [id] => 1
                                    [name] => Alkoholgehalt
                                    [filterable] => 1
                                )

                            [1] => Array
                                (
                                    [id] => 2
                                    [name] => Flaschengröße
                                    [filterable] => 1
                                )

                            [2] => Array
                                (
                                    [id] => 4
                                    [name] => Farbe
                                    [filterable] => 1
                                )

                            [3] => Array
                                (
                                    [id] => 6
                                    [name] => Geschmack
                                    [filterable] => 1
                                )

                            [4] => Array
                                (
                                    [id] => 7
                                    [name] => Trinktemperatur
                                    [filterable] => 1
                                )

                        )

                    [attribute] => 
                )

        )

    [total] => 1
)
```

vanWittlaer:
OK, this working example

`$properties = array(
     "name" => "Liquor",
     'position' => 1,
     'comparable' => 1,
     'sortmode' => 2
);`

is on https://developers.shopware.com/developers-guide/rest-api/examples/article/#example-6-product-properties so it appears spelling for getList and create is different? I admit I did not try w/ getList but rather with create only - and there it should be 'sortmode'. Kind of weired to me. Should we correct the doc for the POST?

mitelg:
I don't understand why this has to be written in lowercase. because the [create method](https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Components/Api/Resource/PropertyGroup.php#L116) uses the [`fromArray` method ](https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Components/Model/ModelEntity.php#L47) which looks up for setter methods. and the setter for the `sortMode` is [`setSortMode`](https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Models/Property/Group.php#L206).
that why it has to be in camelCase, imo :blush:

mitelg:
ok, the Problem is this method: https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Components/Api/Resource/PropertyGroup.php#L215
and somehow `setSortmode` seems to work, even if the method name is not correct at all

vanWittlaer:
My point is sortMode does not (!) work for create and I'd take Shopware/Components/Api/Resource/PropertyGroup.php:215 rather as a confirmation of this than being the cause of it.

mitelg:
yes, we should adjust the create example. but imo this is a bug in `preparePropertyData` method :laughing: 
I just wanted to state out, what the problem is. 
I will reopen your PR. could you revert the GET example?

vanWittlaer:
OK as such?

mitelg:
:top: thanks a lot @vanWittlaer :+1: :tada: :blue_heart:

PR #870, opened 2018-06-18
Add some active flags

t2oh4e:
The active = true flag of the mainDetail is essential because otherwise the article will show up as active in the backend but not in the frontend.


mitelg:
thanks @t2oh4e :+1: :tada: :blue_heart:

PR #876, opened 2018-07-03
Small typo fix

jhit:
German `und` needs to be `and` in English


mitelg:
good catch :+1: :tada: :blue_heart:

PR #881, opened 2018-07-20
InstallContext instead of ...Context

oktupol:
Looking at `$context->scheduleClearCache(...Context::CACHE_LIST_ALL);` I thought `scheduleClearCache` takes a variable number of arguments instead of an array. This commit clears that up


mitelg:
thanks @oktupol :+1: :tada: :blue_heart:

PR #895, opened 2018-08-22
change directory structure

mnaczenski:
The directory structure is wrong, since the document folder is a subfolder of the Bare theme and not the responsive theme.


mitelg:
good catch! thanks @mnaczenski :+1: :tada: :blue_heart:

PR #910, opened 2018-10-02
Add example for DELETE (Stack)

Pmuckel:
Added info that it's possible to delete articles via article number as well
Added example body for batchDelete Request


source/developers-guide/rest-api/api-resource-article/index.md
```diff
@@ -213,12 +213,26 @@ The article(s) to delete can be defined using the following syntax:
 
 ## DELETE (Stack)
 
-In order to delete more than one article at once, it's possible to provide an array of ids to the REST API.
-Simply pass the array of article ids to the following URL (example)
+In order to delete more than one article at once, it's possible to provide an array of objects with ids or article numbers to the REST API.
```
mitelg: I know it is the "article" resource, but it hurts my eyes, when I'm reading of `articles` :smile: 
I think we should use the correct translation `product` if we are talking about non-code-related stuff

what do you think about that @soebbing ?

soebbing: I agree, but we should do so completely. Changing it only here would be suprising.

soebbing: I made a PR which changes "article" to "product" where i thought it would fit

mitelg: @soebbing I merged your PR.

@Pmuckel please rebase your branch :blush:

Pmuckel: @mitelg done

mitelg:
thanks for your contribution @Pmuckel :+1: :tada: :blue_heart:

PR #924, opened 2018-10-17
Added hints about the cronjob.xml

JoshuaBehrens:
As I ran into an issue that customized cronjobs intervals where set back on a plugin update I searched for the reason, found it and now writing a hint for the documentation so anyone who will discover the `cronjob.xml` for their first time will have some "behind the scenes" knowledge about it.


mitelg:
thanks for your contribution @JoshuaBehrens 👍🎉💙

PR #927, opened 2018-10-22
Updates info about uninstalling a plugin

jkrzefski:
With the new plugin system the question if user data should be removed will always be asked. It is not optional anymore and cannot be invoked manually.


mitelg:
thanks for your contribution @jkrzefski :+1: :blue_heart:

PR #928, opened 2018-10-24
Corrected type of event arg for doctrine events

buddhaCode:
+ Distinguish between Doctine's and Shopware's event system
+ Correct example. In fact the given events args parameter is not of type LifecycleEventArgs, but of type \Enlight_Event_EventArgs. The Doctrine event is wrapped into the shopware event. Pls, see: https://github.com/shopware/shopware/blob/09fb7841a02a86631c00b13c8d1b8e99b9bc38b2/engine/Shopware/Components/Model/EventSubscriber.php#L194

### DIFF
--- a/source/developers-guide/models/index.md
+++ b/source/developers-guide/models/index.md
@@ -77,7 +77,7 @@ _The name of the model is always on the left side of the model, with which the a
 
 In addition to the Shopware events, Doctrine also offers an event system. This event system has been implemented into Shopware to add several new features. Just as like in Shopware, there are the `pre*` and `post*` events. Naturally, the `pre*` events fire before the respective action, and the `post*` events afterwards.
 
-To find more in depth documentation about Doctrine events, please refer to the official [Doctrine documentation](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/events.html).
+You can either define the events on the model class itself (using Doctrine's event system) or in a seperate subscriber class (using Shopware's event system). To find more in depth documentation about Doctrine events and defining the events on the model class, please refer to the official [Doctrine documentation](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/events.html). The following examples refers to defining the events in a seperate subscriber class.
 
 ### Insert event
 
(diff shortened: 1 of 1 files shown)


source/developers-guide/models/index.md
```diff
@@ -104,8 +103,8 @@ class ModelSubscriber implements EventSubscriber
     public function getSubscribedEvents()
     {
         return [
-           Events::prePersist,
-           Events::postPersist,
+           'Article::prePersist',
```
mitelg: I think both are valid. see https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Models/Media/MediaSubscriber.php#L60

and then we should get the `LifecycleEventArgs`

source/developers-guide/models/index.md
```diff
@@ -89,7 +89,6 @@ When a model object is first added to the database, the `prePersist` and `postPe
 namespace SwagModel\Subscriber;
 
 use Doctrine\Common\EventSubscriber;
```
mitelg: we think that our example is correct, as we are using the `Doctrine\Common\EventSubscriber` and not the Shopware event subscriber. 
maybe you can add another example with the Shopware Eventsubscriber?

source/developers-guide/models/index.md
buddhaCode: Yes, that's totally fine. You get the LifecycleEventArgs when defining the handlers direct in the model class. In that case you are using the the original doctrine event system. The MediaSubscriber implements the Doctrine\Common\EventSubscriber interface. This is the same as defining them in directly in the model class and, again, belongs to the doctrine event system. (https://github.com/shopware/shopware/blob/4fb20a661576322704ea0d20332bdbe3c49e2ab6/engine/Shopware/Components/DependencyInjection/services.xml#L308)

But in this example you are defining the event handlers in a shopware subscriber class. In this case you using shopware's event system. Doing so, you will get the shopware event class and not the doctrine event class.

Registering Events::prePersist to the event model doesn't fit to rest of the example. Within the handler the example refers to the article model. Thats the reason I changed Events to Article.

buddhaCode: Sorry. guess you are right. I didn't see the line '<tag name="doctrine.event_subscriber"/>' in the example. In my case I was using '<tag name="shopware.event_subscriber" />'. Which works to, but you will get the Enlight_Event_EventArgs instead of LifecycleEventArgs.

source/developers-guide/models/index.md
buddhaCode: Sure. But I think there is no need, isn't it?

PR #931, opened 2018-10-29
Update php-cs-fixer call

nlubisch:
I updated the php-cs-fixer call because the option "level" doesn't exist and the general `.php_cs.dist` should be considered.


mitelg:
thanks for your contribution @nlubisch :+1: :blue_heart: :tada:

PR #932, opened 2018-10-30
Updated POST and PUT for categories

Pmuckel:
Updated the available fields for categories' POST and PUT requests and added an example for a POST request


mitelg:
thanks @Pmuckel :+1: :blue_heart:

PR #939, opened 2018-11-09
[FEATURE] Added new documentation on how to prepare themes for the ne…

wolv-dev:
…w plugin system


source/designers-guide/preparing-themes-for-the-community-store/index.md
```diff
@@ -25,24 +25,87 @@ The plugin directory has to have a specific structure in order to work inside Sh
...
+{
+}
+```
+
+You need to create a `plugin.xml` file in the same directory, which defines the meta data of your plugin, i.e. label, version, compatibility or the changelog.
+
+##### plugin.xml
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<plugin xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/shopware/5.5/engine/Shopware/Components/Plugin/schema/plugin.xsd">
```
mitelg: if the minVersion of the plugin is SW 5.2.0 we usually use the same xsd path here. so the path to the 5.2 version

source/designers-guide/preparing-themes-for-the-community-store/index.md
```diff
@@ -25,24 +25,87 @@ The plugin directory has to have a specific structure in order to work inside Sh
...
+    <version>1.0.0</version>
+    <copyright>(c) by shopware AG</copyright>
+    <license>MIT</license>
+    <link>http://store.shopware.com</link>
+    <author>shopware AG</author>
+    <compatibility minVersion="5.2.0"/>
+</plugin>
+```
+
+The plugin should now be displayed inside the Shopware plugin manager, where it can be installed. Once the plugin is enabled, the theme will be available and selectable in the theme manager, along with all other existing themes. If this is the case, the plugin is ready to be published in the Shopware Community Store.
```
mitelg: maybe you should say `activated` instead of `enabled`

mitelg:
nice work @wolv-dev !
one thing: could you provide the example plugins inside the [exampleplugins](https://github.com/shopware/devdocs/tree/master/exampleplugins) directory? so in this way, one could easily adjust the plugin if necessary

wolv-dev:
@mitelg I just renamed the existing plugin to "_legacy" and created the new one in the same directory. Should I move the legacy plugin as well?

wolv-dev:
@mitelg I added both plugins into the exampleplugins directory and made the annotated changes

mitelg:
one more thing :blush: 
you could remove the zip files and add links like in this example: https://github.com/shopware/devdocs/blob/master/source/developers-guide/backend-extension/index.md

the zip files for the plugins are generated automatically

mitelg:
thanks again for your great work @wolv-dev :+1: :blue_heart: 
keep on the good work :blush:

PR #940, opened 2018-11-10
Fix value of unitize mixin property

pr-olga:
Property-value relationship in unitze mixin is defined as `@{property}: ~"@{pxValue}px"`. So the `px` ending in mixin causes  an invalid value.


mitelg:
good catch @pr-olga thanks for your contribution :+1: :tada: :blue_heart:

PR #944, opened 2018-11-14
Change path of install.sh

Pmuckel:
Updated the path of the install.sh like it is on https://github.com/shopware/composer-project#installation


mitelg:
thanks @Pmuckel :+1:

PR #951, opened 2018-11-23
NTR - add elasticsearch version

PascalThesing:
Add the documentation for the new elasticsearch version config.

### DIFF
--- a/source/sysadmins-guide/elasticsearch-setup/index.md
+++ b/source/sysadmins-guide/elasticsearch-setup/index.md
@@ -52,6 +52,7 @@ To enable Elasticsearch (provided it's already installed, configured and running
         'enabled' => true,
         'number_of_replicas' => null,
         'number_of_shards' => null,
+        'version' => '5.6.5'
         'client' => [
             'hosts' => [
                 'localhost:9200'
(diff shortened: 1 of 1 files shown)


source/sysadmins-guide/elasticsearch-setup/index.md
```diff
@@ -87,7 +89,7 @@ return [
 ];
 ```
 
-Shopware 5 communicates with Elasticsearch using the latter's REST API. The `hosts` array accepts multiple address syntaxes, about which you can read more [here](https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/_configuration.html#_host_configuration). The `number_of_shards` and `number_of_replicas` parameter provided to the generated index. A `null` configuration allows to use the elastic search server configuration of this parameters.
+Shopware 5 communicates with Elasticsearch using the latter's REST API. The `hosts` array accepts multiple address syntaxes, about which you can read more [here](https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/_configuration.html#_host_configuration). The version defines the used Elasticsearch Version. If not set in the config, Shopware will detect it automatically. The `number_of_shards` and `number_of_replicas` parameter provided to the generated index. A `null` configuration allows to use the elastic search server configuration of this parameters.
```
mitelg: maybe you should state here, why should one set the version in the config -> performance, etc

mitelg:
so this is ok @soebbing ?

PascalThesing:
I think so @mitelg.

mitelg:
thanks @PascalThesing :blush:

PR #952, opened 2018-11-23
Update plugin for custom productboxlayout to work with infinite scrol…

Pmuckel:
…ling

The template directory for the custom productboxlayout-template hasn't been known to the widget's view as the plugin only subscribed to the frontend controller and not to the widgets controller which is responsible for infinite scrolling.


Pmuckel:
I moved the plugin to the new directory and edited the link inside the index.md.

exampleplugins/5.2/SwagCustomProductBoxLayout/Subscriber/Frontend.php
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public function __construct($pluginDirectory)
+    {
+        $this->pluginDirectory = $pluginDirectory;
+    }
+    
+    public static function getSubscribedEvents()
+    {
+        return array(
+            'Enlight_Controller_Action_PostDispatchSecure_Frontend_Listing' => 'onFrontendPostDispatchListing',
```
mitelg: if you only use these two events, your template directory will only be loaded on listing pages. this could be a problem, if the product boxes are used in sliders, e.g. on the detail page. did you check that? maybe you should use something like this:
```
    public static function getSubscribedEvents()
    {
        return [
            'Theme_Inheritance_Template_Directories_Collected' => 'onTemplateDirectoriesCollect',
        ];
    }

    public function onTemplateDirectoriesCollect(EventArgs $args)
    {
        $dirs = $args->getReturn();
        $dirs[] = $this->pluginPath . '/Resources/views/';
        $args->setReturn($dirs);
    }
```

exampleplugins/5.2/SwagCustomProductBoxLayout/SwagCustomProductBoxLayout.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+class SwagCustomProductBoxLayout extends Plugin
+{
+
+    /**
+    * @param ContainerBuilder $container
+    */
+    public function build(ContainerBuilder $container)
+    {
+        $container->setParameter('swag_custom_product_box_layout.plugin_dir', $this->getPath());
```
mitelg: this is not needed since SW 5.2.xsomething. so you can remove the build method, as we have 5.5.3 now :wink:

exampleplugins/5.2/SwagCustomProductBoxLayout/plugin.xml
```diff
@@ -0,0 +1,18 @@
...
+        xsi:noNamespaceSchemaLocation="../../../engine/Shopware/Components/Plugin/schema/plugin.xsd">
+    <label lang="de">SwagCustomProductBoxLayout</label>
+    <label lang="en">SwagCustomProductBoxLayout</label>
+    <version>1.0.0</version>
+    <copyright>(c) by Shopware AG</copyright>
+    <license>proprietary</license>
+    <link>http://store.shopware.com</link>
+    <author>Shopware AG</author>
+
+    <compatibility minVersion="5.2.0" />
```
mitelg: ok, did not see the min version :smile: 

than if you set this up, you can remove the build method from the base class

exampleplugins/5.2/SwagCustomProductBoxLayout/Subscriber/Frontend.php
mitelg: if you use the suggested event, the addTemplateDir in the backend subscriber should be obsolete, too

mitelg: maybe @Pmuckel could add this event? :see_no_evil:

exampleplugins/5.2/SwagCustomProductBoxLayout/plugin.xml
Pmuckel: I updated the minVersion to 5.2.10 and removed the build method.

mitelg:
thank you very muck @Pmuckel :+1: :blue_heart:

PR #955, opened 2018-11-27
Add Theme_Inheritance_Template_Directories_Collected event

Pmuckel:


source/developers-guide/event-list/index.md
```diff
@@ -1354,3 +1354,65 @@ stdClass Object
...
+### Listener
+```
+public function onCollectTemplateDirectories(\Enlight_Event_EventArgs $args)
+{
+    /** @var $directories array */
+    $directories = $args->getReturn();
+
+    /** @var $template \Shopware\Models\Shop\Template */
+    $template = $args->get('template');
+
```
mitelg: a short example, how to set a new directory would be nice. because it is not similar to the known `addTemplateDir` thing

PR #967, opened 2019-01-08
Removed margin-top from .code-collapsed

pascalheidman-bedarf:
Only on collapsed code there is a margin on top resulting in ugly jumping of code block.


mitelg:
thanks for your contribution @pascalheidman-bedarf  :+1: :blue_heart:

PR #971, opened 2019-01-11
Add note for own single & multiselection attributes

nlubisch:


mitelg:
thanks @nlubisch for pointing that out! :+1: :blue_heart:

PR #976, opened 2019-01-15
Important information was missing here

snHolst:
Missing use information was added in the text.


mitelg:
hey @snHolst 
thanks for your contribution, but missing `use` statements is something that your IDE should take care off.

PR #977, opened 2019-01-16
SW-23264 - add dynamic mapping

PascalThesing:


source/sysadmins-guide/elasticsearch-setup/index.md
```diff
@@ -112,6 +114,32 @@ return [
 ];
 ```
 
+
+### Define Elasticsaerch Dynamic Mapping
+
+By default, Shopware enables the dynamic mapping option.
+With this fields can be added dynamically to a document or to inner objects within a document, just by indexing a document containing the new field.
```
mitelg: By this, fields can be added ... 

the comma is important

source/sysadmins-guide/elasticsearch-setup/index.md
```diff
@@ -112,6 +114,32 @@ return [
...
+
+The problem with this option is, that Shopware provides floats and integers to ElasticSearch.
```
mitelg: you should write `Elasticsearch` as this is the proper name. also in the following text

PR #978, opened 2019-01-17
Fix wrong config example

mnaczenski:
The example won't work because of a syntax error.


mitelg:
thanks @mnaczenski :+1: :blue_heart:

PR #982, opened 2019-01-29
Added reference to existing datepicker documentation

JoshuaBehrens:
It was weird to neither have a preview nor a code example. The reference to the jquery ui page said: `$('input').datepicker();` which was not working in shopware and is therfore confusing. As the current js files in styletile do not contain the datepicker I did not include a preview iframe but referenced the documention for an approach that is a bit more clear.


mitelg:
thanks for your contribution @JoshuaBehrens :+1: :blue_heart:

PR #985, opened 2019-02-05
Using a working example to display a product name

JoshuaBehrens:
One of our apprentices noticed an error while working through the documentation :smirk: The given example will output nothing and confuses newbies.

I am not sure about the post about the seo engine that might have the same issue: https://github.com/shopware/devdocs/blob/93cb1789f9b93acf88d82d545ae0883d21e892eb/source/blog/_posts/2015-08-11-seo-engine.md#configuring-the-urls
There it also uses `sArticle.name` 
> `{sCategoryPath articleID=$sArticle.id}/{$sArticle.id}/{$sArticle.name}`

### DIFF
--- a/source/designers-guide/smarty/index.md
+++ b/source/designers-guide/smarty/index.md
@@ -39,7 +39,7 @@ To start with, we use Smarty to output the data we receive from the shop system.
 **Example:** *Variable output*
 
 ```html
-<h2>{$sArticle.name}</h2>
+<h2>{$sArticle.articleName}</h2>
 ```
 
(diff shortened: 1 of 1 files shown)


mitelg:
thanks for the fix @JoshuaBehrens :+1: :blue_heart: 
If the SEO example should be wrong, another PR is appreciated :blush:

JoshuaBehrens:
I just had a look at the default configuration 
image and this looks fine for me then. Although I am confused why this works :confused:

oktupol:
@JoshuaBehrens It works because of this:
https://github.com/shopware/shopware/blob/5644207b68a5ed430227b9308662bfaf8796ab02/engine/Shopware/Core/sRewriteTable.php#L850-L857

JoshuaBehrens:
Thanks for the explanation

PR #989, opened 2019-02-22
Fix wrong filter description in rest api

mnaczenski:
See: https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Components/Api/Resource/Article.php#L221-L226

The filter ?language uses the shopID, not a locale. Since translation of entities are based on the shop, not a locale, it won't make any sence to use the locale here. Further the locale is not a unique alias, since more than one shop can have the locale en_GB.


mitelg:
thanks @mnaczenski :+1: :blue_heart:

PR #993, opened 2019-02-26
Explicitly mention $Data

EtienneBruines:
To avoid confusion, it should be good to mention the data appearing under 
$Data.key instead of $key directly.


mitelg:
thanks for your PR @EtienneBruines :+1: :blue_heart:

PR #994, opened 2019-03-06
Added missing first line to build the javascript correctly

JoshuaBehrens:
As this seems to be a :clipboard: :ramen: guide the files should contain out of the box correct content to make it easier to get started with ext js. The first line in the included files should be added with useless statements or smarty statements as in the compiled main.js this line is commented out.
If this first line is missing invalid javascript is generated.


mitelg:
thanks @JoshuaBehrens :+1: :blue_heart:

PR #999, opened 2019-03-20
Fix typo

rschweizer:
Following JSON syntax, `result` must be an object in this example, rather than an array.


mitelg:
good catch :+1: thanks for your first contribution :blush: :tada: :blue_heart:

PR #1001, opened 2019-03-27
change integer to boolean

tinect:
because it should be boolean


mitelg:
thanks @tinect :+1: :tada: :blue_heart:

PR #1005, opened 2019-05-16
Fix path to global tests dir

mattsches:


mitelg:
thanks for your contribution @mattsches :+1: :tada: :blue_heart:

PR #1006, opened 2019-05-17
NTR - add hook interface

PascalThesing:


mitelg:
thanks @PascalThesing :blue_heart:

PR #1009, opened 2019-06-06
Add advice to remove the engine directory

larsbo:


source/blog/_posts/2018-07-25-switching-to-composer-installation.md
```diff
@@ -49,7 +49,7 @@ mv composer-installation/{app,bin,Plugins,vendor,composer.json,shopware.php} ./
 mv composer-installation/.env.example ./.env
 ```
 
-The directories `custom`, `files`, `media`, `themes`, `var` and `web` are equal with the composer installation so we can still use them and don't lose our files, plugins and themes. If you have plugins in `engine/Shopware/Plugins/Community` or `engine/Shopware/Plugins/Local` they must be moved to ```./Plugins```!
+The directories `custom`, `files`, `media`, `themes`, `var` and `web` are equal with the composer installation so we can still use them and don't lose our files, plugins and themes. If you have plugins in `engine/Shopware/Plugins/Community` or `engine/Shopware/Plugins/Local` they must be moved to ```./Plugins```! Now we can delete the old `engine` directory because from now on the Shopware core files will be provided by the `shopware/shopware` composer package inside the `vendor` directory.
```
shyim: There is missing then the part of the symlinking https://github.com/shopware/composer-project/blob/master/app/bin/functions.sh#L171-L189

larsbo: avoid the symlinks 😉 
What about running `post-install.sh` afterwards? It recreates the engine folder with the symlinks.

shyim: It's currently does that, but when we change there something this Guide will be broken then :)

larsbo: That's true. But there is no guarantee that changes in the composer project will _not_ resulting in breaking this guide without this change 😉 
What bothers me is that at the moment this guide doesn't create a clean composer project at all. Instead of the Symlinks there will remain the original files in the `engine` folder which results in a lot of duplicated files.

mitelg:
thanks for your contribution @larsbo :+1: :tada: :blue_heart:

PR #1018, opened 2019-06-17
make it possible to set `noindex, nofollow` for blog entries

ndzoesch:


source/_layouts/default.twig.html
```diff
@@ -12,7 +12,9 @@
     {% endblock %}
 
     {% if page.menu_chapter %}
```
janbuecker: Hmm.. I think that this IF is quite confusing as you don't check on `robots.hide`, only in the else block.

ndzoesch: I know what you mean. The if was there before, and I did not want to risk having two meta tags with the same content. Any idea how to solve this more elegant? @janbuecker

janbuecker: What about this?

```
{% if page.menu_chapter %}
    {% if page.robots.hide == true %}
        <meta name="robots" content="noindex, nofollow" />
    {% else %}
        <meta name="robots" content="noindex" />
    {% endif %}
{% endif %}
```

ndzoesch: Won't work, since blog post do not have the page.menu_chapter property. 😐

PR #1022, opened 2019-07-16
Amend configuration key

JoshuaBehrens:
See the original key: https://github.com/shopware/shopware/blob/354cb81eb3ed11b6cad8e2861aefc4a3796c1e9f/engine/Shopware/Configs/Default.php#L157


mitelg:
good catch!
thanks for your contribution @JoshuaBehrens :+1: :tada: :blue_heart:

PR #1027, opened 2019-08-16
wrong http method

RobTranquillo:
change http method to post and use a more common roleId


mitelg:
thanks for your contribution @RobTranquillo :+1: :tada: :blue_heart:

PR #1036, opened 2019-10-01
Fix Order Shipping Model

nlubisch:
The documentation and the given result of an order request differs so i fixed in this part the order shipping models


mitelg:

PR #1037, opened 2019-10-01
Fix document type model

nlubisch:
The previous model doesn't exist so fixed this to the correct one.


mitelg:

PR #1038, opened 2019-10-01
Added payment status

nlubisch:
The payment status link on https://developers.shopware.com/developers-guide/rest-api/api-resource-orders/ leads to nothing so i added the payment status model.
I also fixed the column "description" to "name" as this is the naming in the api result.


mitelg:

PR #1044, opened 2019-10-09
Rewrite API examples to JSON, add badges to represent the HTTP request method

philipgatzka:
Like described in the title - I rewrote all examples from using the example-client written in PHP to JSON. This means the examples can now be copied and executed with any HTTP-Client.

Instead of showing the method to be used via function calls `$client->post()`, they're now displayed using a colour-coded badge.

There are two issues still left to be discussed I think:
1. Should the example client still be documented? (I left it as it is)
2. There are a few pages which show `GET` requests [with request-bodies](https://developers.shopware.com/developers-guide/rest-api/customer-streams/#get-information-about-an-existing-stream) and I didn't rewrite those, as I think they rely on non-standard behaviour. Should these examples be kept or deleted?

### DIFF
--- a/source/_partials/api_badge.twig
+++ b/source/_partials/api_badge.twig
@@ -1,0 +1,8 @@
+<div class="api-badge api-badge-{{ method|lower }} {% if body %}body{% endif %}">
+  <span class="method">
+    {{ method }}
+  </span>
+  <span class="route">
+    {{ route|raw }}
+  </span>
+</div>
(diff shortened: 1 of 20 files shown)


mitelg:
looks good :+1: 
1. I would delete the example client. JSON is not restricted to any programming language and everybody knows how to use it, in his environment. 
2. Especially cases which are not standard must be documented! so of course do not delete them

philipgatzka:
Yes, that sounds reasonable :sweat_smile: I removed the PHP example client and documented the other cases I left out before.

mitelg:
thank you very much @philipgatzka :+1: :tada: :blue_heart: 
really nice work

PR #1050, opened 2019-10-24
Add condition when plugin is not active

shyim:


mitelg:
thanks @shyim :+1: :blue_heart:

PR #1051, opened 2019-10-28
Remove file example

shyim:
File is not a valid session handler on cli
PHP Warning:  ini_set(): Cannot find save handler 'file' in /var/www/html/shopware/engine/Library/Zend/Session.php on line 213


mitelg:
thanks @shyim :+1:

mitelg:
btw PR No. 1000 :tada: :tada: :birthday:

PR #1052, opened 2019-10-28
Add missing required parentId

shyim:
Fix examples :)
```
Fatal error: Uncaught Error: Call to a member function getId() on null in /var/www/html/shopware/engine/Shopware/Components/Model/CategorySubscriber.php:296
```


mitelg:
thanks @shyim :+1:

PR #1073, opened 2020-05-19
Fix broken SwagController example plugin

PaddyS:
Wrong namespace is used in the SwagController example


mitelg:
thanks @PaddyS :+1:

PR #1089, opened 2020-10-01
Improved Docs

ss9125570:


mitelg:
hey @ss9125570 
we will close this PR, as it does not really add value to the project. Also it seems to be so called hacktoberfest spam. If this is not the case, we are very sorry, and will appreciate a new PR once October is over :wink:

PR #1091, opened 2020-10-05
Fix article related and similar column

nlubisch:
I don't want to be a shitty hacktoberfest developer but this is something i stumbled upon this day 😄 

The field "name" is not used in this context - the source code revealed that it should be "number".


mitelg:
aren't these the IDs of the related/similar products? :thinking: I know the column is a `VARCHAR` :see_no_evil:  but the values do not look like the product number to me

nlubisch:
I found another more complex problem. For writing new related/similars you need to pass the field "number" with the articleNumber (works for us) but when you read them you get the field "name" with the name of the article. 

It's a bit misleading in the docs as both behaviours are not visible on the first sight.

nlubisch:
I've added a seperate section for get and put/post for both models to represent the correct behaviour.

mitelg (approved):

mitelg:
thanks for your contribution @nlubisch  :+1: :tada: :blue_heart:

PR #1093, opened 2020-10-18
Update index.html

nagpalnipun22:
Read the head section and metadata
Edit the body, title and HTML tags

### DIFF
--- a/source/index.html
+++ b/source/index.html
@@ -7,6 +7,15 @@ pagination:
 use:
 - posts
 ---
+<head>
+</head>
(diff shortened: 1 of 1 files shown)


mitelg:
hello @nagpalnipun22 
thanks for your contribution, but the meta data is configured in this file https://github.com/shopware/devdocs/blob/master/source/_layouts/start.twig.html

so there is no need to add it here. also have a look at the homepage https://developers.shopware.com
as you can see here, there is nothing missing :blush:

PR #1099, opened 2021-04-30
corrected doc for elastic search

Abbna:
corrected doc for addressing of newly created mapping for elastic search, as the properties attributes have to be ignored when addressing the field

I had massive problems to address attributes created over the free text management inside the elastic search queries generated by using the ONGR Framework.
I noticed, that this .properties is too much. It's an attribute inside the mapping, but it does not work to address the field with it. By addressing it has to be removed.


shyim (approved):
Looks good

PR #1105, opened 2021-09-26
fix location of cache and logs folder

jonashrem:
cache and logs folder are both inside "var"


philipgatzka:
Hey @jonashrem , thank you for your contribution! 🎉

PR #1121, opened 2022-02-17
Update `shipped` data type

erikforss:
The data type of the shipped value is wrong in the developers documentation.
The data type should be integer:
https://github.com/shopware/shopware/blob/c8558e5da009b1b608d77f1decff6484bbe52cf0/engine/Shopware/Models/Order/Detail.php#L206-L211


mitelg:
thanks for your contribution @erikforss   👍🎉💙

PR #1127, opened 2023-01-09
SW-25113 - update symfony version

PascalThesing:


composer.json
```diff
@@ -1,18 +1,18 @@
...
         "league/commonmark": "^1.3.0",
-        "mavimo/sculpin-redirect-bundle": "@dev",
-        "sculpin/sculpin": "2.1.2",
-        "symfony/config": "2.8.52",
-        "symfony/dependency-injection": "2.8.52",
-        "symfony/event-dispatcher": "2.8.52",
-        "symfony/http-kernel": "2.8.52",
-        "twig/twig": "1.44.7"
+        "sculpin/sculpin": "3.2.0",
+        "symfony/config": "^4.4.13",
```
mitelg, line 10: why not using latest Symfony?

source/community/index.html
```diff
@@ -6,8 +6,6 @@
 menu_title: Community
 menu_order: 35
 menu_style: bullet
-redirect:
```
mitelg, line 9: why did you removed the redirects? this will cause breaks for old links, I guess

mitelg (changes requested):

composer.json
PascalThesing: This is the requirement of sculin symfony versions. 
They should match.

source/community/index.html
PascalThesing: The sculpin redirect link bundle is broken with newer versions and since this is not maintained by someone which we knew and these are just a few links, I decided to remove these few.

PR #1133, opened 2023-11-09
fix: broken download and install guide links

mario-neuhold:
Just about to install Shopware 5 for the first time when I stumbled upon these dead links in the docs. Found some working ones that might be a better fit.

Download from git release page: https://github.com/shopware5/shopware/releases
Installation instructions: https://docs.shopware.com/en/shopware-5-en/first-steps/installing-shopware?category=shopware-5-en/getting-started

Also the current official download page doesn't have Shopware 5 anymore, so the install guide should get updated as well:
image


mitelg (approved):
thanks for your contribution @mario-neuhold 👍🎉💙

I will take care of our user documentation. thanks for pointing out

